PR #2, opened 2021-04-06
fix: README zebra link pointing at zcash

Mirko-von-Leipzig:
Both Zebra and ZCash links were pointing at the ZCash repo. This updates the Zebra link to point at the Zebra repo.


niklaslong (approved):
Thanks! 🚀

PR #3, opened 2021-04-09
feat/message filter

Mirko-von-Leipzig:
Adds the `MessageFilter` and the `unsolicitation_listener` test case.

I added quite a bit of docs to the `MessageFilter` - mostly as an exercise to learn Rust docs.

### DIFF
--- a/src/protocol/message.rs
+++ b/src/protocol/message.rs
@@ -180,3 +180,183 @@ fn checksum(bytes: &[u8]) -> u32 {
 
     u32::from_le_bytes(checksum)
 }
+
+/// Controls the filter response of [MessageFilter] to messages it reads.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Filter {
+    /// Do not filter message
+    Disabled,
+    /// Filter message
+    Enabled,
+    /// Filter message and reply with a default response
+    AutoReply,
+}
+
+/// Provides a wrapper around [Message::read_from_stream] which optionally filters
+/// certain Message types, and can send default responses if required.
+///
+/// This can be used to wait for a Message event that you actually care about,
+/// while skipping over spurious Message requests e.g. [Message::Ping].
+///
+/// Currently supports filters for the following [Message] types:
+///     - [Message::Ping]
+///     - [Message::GetHeaders]
+///     - [Message::GetAddr]
+///     - [Message::GetData]
+///
+/// For a list of responses see the documentation on [MessageFilter::read_from_stream].
+///
+/// Can optionally log filter events to console, logging is disabled by default.
+pub struct MessageFilter {
+    ping: Filter,
+    getheaders: Filter,
+    getaddr: Filter,
+    getdata: Filter,
+    // todo: inv
+    // todo: getblocks
+    // todo: mempool
+    logging: bool,
+}
+
+impl MessageFilter {
+    /// Constructs a [MessageFilter] which will filter no messages, and with logging disabled.
+    pub fn with_all_disabled() -> Self {
+        use Filter::Disabled;
+
+        Self {
+            ping: Disabled,
+            getheaders: Disabled,
+            getaddr: Disabled,
+            getdata: Disabled,
+
+            logging: false,
+        }
+    }
+
+    /// Constructs a [MessageFilter] which will filter all supported message types, and with logging disabled.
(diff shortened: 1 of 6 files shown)


src/protocol/message.rs
```diff
@@ -66,6 +66,7 @@ impl MessageHeader {
     }
 }
 
+#[derive(Debug)]
```
niklaslong, line 69: Thanks for adding these! 👍

src/tests/conformance/messages.rs
```diff
@@ -51,3 +51,32 @@ async fn handshake(node_addr: SocketAddr) -> io::Result<TcpStream> {
 
     Ok(peer_stream)
 }
+
+#[tokio::test]
+async fn unsolicitation_listener() {
```
niklaslong, line 56: As this test isn't part of the index, we may want to ignore it by default (with `#[ignore]`)?

src/protocol/message.rs
```diff
@@ -179,3 +180,187 @@ fn checksum(bytes: &[u8]) -> u32 {
...
+            Message::GetHeaders(_) => {
+                Message::Headers(Headers::empty())
+                    .write_to_stream(stream)
+                    .await
+            }
+            Message::GetData(inv) => Message::NotFound(inv).write_to_stream(stream).await,
+            _ => unimplemented!(),
+        };
+
+        result
```
niklaslong, line 308: Suggestion: instead assigning the result, you could use `?` to bubble the error and return `Ok(())` if the execution isn't short-circuited.

src/protocol/message.rs
```diff
@@ -179,3 +180,187 @@ fn checksum(bytes: &[u8]) -> u32 {
 
     u32::from_le_bytes(checksum)
 }
+
+/// Controls the filter response of [MessageFilter] to messages it reads.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Filter {
```
niklaslong, line 186: Future consideration: we may want to move this code to another directory as it isn't explicitely part ot the protocol. Maybe `src/setup/message_filtering.rs` or similar? We'll also have some handshake helpers there and in future the synthetic nodes.

src/protocol/message.rs
```diff
@@ -179,3 +180,187 @@ fn checksum(bytes: &[u8]) -> u32 {
...
+    pub fn with_ping_filter(mut self, filter: Filter) -> Self {
+        self.ping = filter;
+        self
+    }
+
+    // sends an appropriate reply in response to the received message
+    async fn reply(&self, stream: &mut TcpStream, message: Message) -> io::Result<()> {
+        let result = match message {
+            Message::Ping(nonce) => Message::Pong(nonce).write_to_stream(stream).await,
+            Message::GetAddr => Message::Addr(Addr::empty()).write_to_stream(stream).await,
```
niklaslong, line 298: I think this is fine for now but we'll want more flexibility in future (e.g. returning the addresses a synthetic node is connected to in the mesh for load testing).

src/protocol/payload/block.rs
```diff
@@ -46,12 +47,20 @@ impl LocatorHashes {
...
 }
 
+#[derive(Debug)]
 pub struct Headers {
     count: VarInt,
     headers: Vec<Header>,
 }
 
 impl Headers {
+    pub fn empty() -> Self {
```
niklaslong, line 57: I like how explicit this is, we may want to add `Default` implementations in future if this is a pattern we find ourselves using a lot. Just something to keep in mind.

niklaslong (commented):
This looks great, I left some suggestions and a few more general comments/things to consider in future. 

The documentation is also great, thanks for being so thorough! 🚀

src/protocol/message.rs
Mirko-von-Leipzig: Yes, I was quite uncertain about where to place it. Considered `src/setup/message_filtering.rs`, or creating a more utility focused module. Or even a module for generating fake messages which we may need in the future.

For now I just wanted to keep it compact while we figure it out organically.

src/protocol/message.rs
Mirko-von-Leipzig: That's what I get for copy pasting a previous attempt 🤦

I can in-fact simply return the result straight from the match.

PR #4, opened 2021-04-12
Transaction encoding and decoding

niklaslong:
This PR implements encoding and decoding for V1 -> V4 transactions (V5 isn't yet stable) as per [the Zcash spec](https://zips.z.cash/protocol/canopy.pdf#consensusfrombitcoin). 

This PR also features very simple round trip unit tests (ignored by default). In future more rigourous testing should be considered.

### DIFF
--- a/src/protocol/payload/tx.rs
+++ b/src/protocol/payload/tx.rs
@@ -7,8 +7,8 @@ pub enum Tx {
     V1(TxV1),
     V2(TxV2),
     V3(TxV3),
-    // V4(TxV4),
-    // Not yet stabalised.
+    V4(TxV4),
+    // Not yet stabilised.
     V5,
 }
 
@@ -30,6 +30,11 @@ impl Tx {
                 buffer.write_all(&(3u32 | 1 << 31).to_le_bytes())?;
                 tx.encode(buffer)?;
             }
+            Tx::V4(tx) => {
+                // The overwintered flag IS set.
+                buffer.write_all(&(4u32 | 1 << 31).to_le_bytes())?;
+                tx.encode(buffer)?;
+            }
             _ => unimplemented!(),
         }
 
@@ -49,6 +54,7 @@ impl Tx {
             (1, false) => Self::V1(TxV1::decode(bytes)?),
             (2, false) => Self::V2(TxV2::decode(bytes)?),
             (3, true) => Self::V3(TxV3::decode(bytes)?),
+            (4, true) => Self::V4(TxV4::decode(bytes)?),
             _ => unimplemented!(),
         };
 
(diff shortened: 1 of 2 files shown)


src/protocol/payload/tx.rs
```diff
@@ -0,0 +1,868 @@
...
+            // Extract transaction version and check if overwinter flag is set.
+            (header & LOW_31_BITS, header >> 31 != 0)
+        };
+
+        let tx = match (version, overwinter) {
+            (1, false) => Self::V1(TxV1::decode(bytes)?),
+            (2, false) => Self::V2(TxV2::decode(bytes)?),
+            (3, true) => Self::V3(TxV3::decode(bytes)?),
+            (4, true) => Self::V4(TxV4::decode(bytes)?),
+            _ => unimplemented!(),
```
Mirko-von-Leipzig, line 61: Should we not return an Error indicating bad payload here instead? I know there is a version 5 on the way, but I wouldn't expect an `io::Result<()>` function to panic on a parsing error.

src/protocol/payload/tx.rs
```diff
@@ -0,0 +1,868 @@
...
+    tx_out_count: VarInt,
+    tx_out: Vec<TxOut>,
+
+    lock_time: u32,
+
+    // BCTV14
+    join_split_count: VarInt,
+    join_split: Vec<JoinSplit>,
+
+    // Only present if the join_split count > 0.
```
Mirko-von-Leipzig, line 140: If there is such explicit coupling between these fields, should they not become its own type to encapsulate that? Probably worth the effort since the optional decoding logic is repeating in every version on-wards.

src/protocol/payload/tx.rs
```diff
@@ -0,0 +1,868 @@
...
+        buffer.write_all(&self.vmacs)?;
+
+        self.zkproof.encode(buffer)?;
+        buffer.write_all(&self.enc_cyphertexts)?;
+
+        Ok(())
+    }
+
+    fn decode_bctv14(bytes: &mut Cursor<&[u8]>) -> io::Result<Self> {
+        // TODO: deduplicate (might require generics).
```
Mirko-von-Leipzig, line 617: Alternatives to generics:
1. add a boolean parameter to select which proof type to use. Not great if you wanted to have everything use the same `decode(bytes)` Trait eventually.
2. figure out which type it is by validating? I presume that's not really feasible or desirable though. 

Not really good ideas, sorry.

niklaslong: > add a boolean parameter to select which proof type to use. Not great if you wanted to have everything use the same decode(bytes) Trait eventually.

Indeed, this is why I avoided it so far. Food for thought, not a pressing matter either way. I had also initially used the remaining length, though that's not really a guarentee we should be relying on.

src/protocol/payload/tx.rs
niklaslong: Great point! 👍 
There are actually quite a few fields that might require explicit types. In the fields I've introduced in this. Given how long this PR is, I was hesitant to include it. I'll jot this down for a follow-up.

Mirko-von-Leipzig (approved):

PR #6, opened 2021-04-13
Improve node startup and handshake

niklaslong:
This PR:
- introduces a `helpers` module to be used for general test heper code until we find a better structure for it,
- relies on peer connection to signal node startup completion (could be further improved, for now this is done manually in tests),
- introduces the `wait_until!` macro—useful for message filtering,
- makes test naming and ordering consistent.

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -1,0 +1,76 @@
+use crate::protocol::{message::Message, payload::Version};
+
+use tokio::{
+    io,
+    net::{TcpListener, TcpStream},
+};
+
+use std::net::SocketAddr;
+
+/// Waits until an expression is true or times out.
+///
+/// Uses polling to cut down on time otherwise used by calling `sleep` in tests.
+#[macro_export]
+macro_rules! wait_until {
+    ($limit_secs: expr, $condition: expr $(, $sleep_millis: expr)?) => {
+        let now = std::time::Instant::now();
+        loop {
+            if $condition {
+                break;
+            }
+
+            // Default timout.
+            let sleep_millis = 10;
+            // Set if present in args.
+            $(let sleep_millis = $sleep_millis;)?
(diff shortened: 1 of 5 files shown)


src/tests/conformance/messages.rs
```diff
@@ -1,55 +1,52 @@
...
-    let verack = Message::read_from_stream(&mut peer_stream).await?;
-    assert!(matches!(verack, Message::Verack));
+    wait_until!(
+        10,
+        matches!(
+            auto_responder
+                .read_from_stream(&mut peer_stream)
+                .await
+                .unwrap(),
+            Message::Pong(..)
```
Mirko-von-Leipzig, line 45: Should include checking if the `nonce` matches?

src/helpers.rs
```diff
@@ -0,0 +1,76 @@
...
+            $(let sleep_millis = $sleep_millis;)?
+            tokio::time::sleep(std::time::Duration::from_millis(sleep_millis)).await;
+            if now.elapsed() > std::time::Duration::from_secs($limit_secs) {
+                panic!("timed out!");
+            }
+        }
+    };
+}
+
+pub async fn handshake(listener: TcpListener) -> io::Result<TcpStream> {
```
Mirko-von-Leipzig, line 34: I find the naming of these guys ambiguous. Maybe `wait_for_handshake`, and `initiate_handshake`? 

I'm not quite convinced by suggestion either. Thoughts?

src/tests/conformance/messages.rs
niklaslong: Good catch!

src/helpers.rs
niklaslong: I'll give it a think (this is likely not the final API), and I'll add a touch of documentation as well.

Mirko-von-Leipzig:
Looks good 👍

PR #8, opened 2021-04-13
refactor: replace nested matches with early exits

Mirko-von-Leipzig:
Removes a lot of the code bloat induced by nested matches. 

Also added a helper function to determine connection error.


niklaslong (approved):
LGTM! 👍

PR #10, opened 2021-04-14
feat/conformance 006

Mirko-von-Leipzig:


src/tests/conformance/handshake.rs
```diff
@@ -275,6 +275,55 @@ async fn reject_non_version_before_handshake() {
...
+
+    Message::Version(
+        Version::new(node.addr(), stream.local_addr().unwrap()).with_nonce(version.nonce()),
+    )
+    .write_to_stream(&mut stream)
+    .await
+    .unwrap();
+
+    // This is required because the zcashd node eagerly sends `ping` and `getheaders` even though
+    // our version message is broken. TBD if this is desired behaviour or if this should fail the test.
```
niklaslong, line 311:
```suggestion
    // our version message is broken. 
    // TODO: tbd if this is desired behaviour or if this should fail the test.
```
So we don't miss this in future ☝️

src/tests/conformance/handshake.rs
```diff
@@ -275,6 +275,55 @@ async fn reject_non_version_before_handshake() {
...
+    let filter = MessageFilter::with_all_disabled()
+        .with_ping_filter(Filter::Enabled)
+        .with_getheaders_filter(Filter::Enabled);
+
+    match filter.read_from_stream(&mut stream).await {
```
niklaslong, line 316: This might benefit from using `wait_until!` (now that it's merged)?

niklaslong (commented):
Great work on this! 🚀  
I made a few small suggestions.

src/protocol/payload/version.rs
```diff
@@ -87,4 +87,13 @@ impl Version {
             relay,
         })
     }
+
+    pub fn nonce(&self) -> Nonce {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I'm happy to do that instead.

src/tests/conformance/handshake.rs
Mirko-von-Leipzig: Good idea 👍

src/tests/conformance/handshake.rs
Mirko-von-Leipzig: I'm not quite following how this would look - is the worry that the read may never occur and we want to timeout?

niklaslong: Indeed, this is more of a nice to have as we would actually notice if the test didn't complete.

niklaslong (approved):

PR #11, opened 2021-04-14
Implement block encoding and decoding

niklaslong:
This PR introduces Block message encoding/decoding and sets up Tx message handling.


Mirko-von-Leipzig:
LGTM

PR #12, opened 2021-04-15
feat: added conformance test 009

Mirko-von-Leipzig:
Tests that a node ignores unsolicited response messages, and does not terminate the connection.

### DIFF
--- a/src/tests/conformance/messages.rs
+++ b/src/tests/conformance/messages.rs
@@ -2,7 +2,7 @@ use crate::{
     helpers::{initiate_handshake, respond_to_handshake},
     protocol::{
         message::{Message, MessageFilter},
-        payload::Nonce,
+        payload::{block::Headers, Addr, Nonce},
     },
     setup::{config::read_config_file, node::Node},
     wait_until,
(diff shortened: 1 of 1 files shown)


src/tests/conformance/messages.rs
```diff
@@ -52,6 +52,63 @@ async fn ping_pong() {
...
+        Message::Ping(nonce)
+            .write_to_stream(&mut stream)
+            .await
+            .unwrap();
+
+        match filter.read_from_stream(&mut stream).await.unwrap() {
+            Message::Pong(returned_nonce) => assert_eq!(nonce, returned_nonce),
+            msg => panic!("Expected pong: {:?}", msg),
+        }
+    }
```
niklaslong, line 109: This is just missing a call to `node.stop().await` (only relevant for docker).

niklaslong (changes requested):
Great work! 🎉

src/tests/conformance/messages.rs
Mirko-von-Leipzig: Thanks for the catch. Should be fixed now.

niklaslong (approved):

PR #13, opened 2021-04-15
feat: add reject message type

Mirko-von-Leipzig:


src/protocol/payload/reject.rs
```diff
@@ -0,0 +1,93 @@
...
+        buffer.write_all(&self.data)
+    }
+
+    pub fn decode(bytes: &mut Cursor<&[u8]>) -> io::Result<Self> {
+        let message = VarStr::decode(bytes)?;
+        let ccode = CCode::decode(bytes)?;
+        let reason = VarStr::decode(bytes)?;
+
+        let mut data = Vec::new();
+        bytes.read_to_end(&mut data)?;
```
niklaslong, line 27: As the length of that field is implied in the body length, reading into a vec with `read_to_end` is probably fine. 

Note: another option would have been to make it a fixed length of 32 bytes ([zebra seems to make that assumption](https://github.com/ZcashFoundation/zebra/blob/24bf952e982bde28eb384b211659159d46150f63/zebra-network/src/protocol/external/codec.rs#L508-L515)).

src/protocol/payload/reject.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+const MALFORMED_CODE: u8 = 0x01;
+const INVALID_CODE: u8 = 0x10;
+const OBSELETE_CODE: u8 = 0x11;
+const DUPLICATE_CODE: u8 = 0x12;
+const NON_STANDARD_CODE: u8 = 0x40;
+const DUST_CODE: u8 = 0x41;
+const INSUFFICIENT_FEE_CODE: u8 = 0x42;
+const CHECKPOINT_CODE: u8 = 0x43;
```
niklaslong, line 45: Zebra seems to have [an extra code](https://github.com/ZcashFoundation/zebra/blob/24bf952e982bde28eb384b211659159d46150f63/zebra-network/src/protocol/external/codec.rs#L504)? Not sure if this is included in zcashd nodes.

niklaslong (commented):
Awesome! 🎉 
We might be missing a ccode but otherwise this looks good to go.

src/protocol/payload/reject.rs
Mirko-von-Leipzig: Its not mentioned in the bitcoin protocol, so might be a zcash addition?

src/protocol/payload/reject.rs
Mirko-von-Leipzig: `Optional extra data provided by some errors. Currently, all errors which provide this field fill it with the TXID or block header hash of the object being rejected, so the field is 32 bytes`

so current behaviour is `Option<[u8; 32]>` -- not sure which we prefer?

niklaslong (approved):

PR #14, opened 2021-04-15
feat: test vectors

niklaslong:
This PR adds blocks for each network upgrade and led to a few tweaks of the en/decoding implementation.


Mirko-von-Leipzig:
LGTM

Mirko-von-Leipzig (approved):
LGTM

PR #15, opened 2021-04-15
feat: check unit tests and lints on CI

niklaslong:
Check Ziggurat unit tests (aka ignored tests), clippy and rustfmt (nightly).


Mirko-von-Leipzig (approved):
LGTM

PR #16, opened 2021-04-16
feat/conformance 007

Mirko-von-Leipzig:
I've added a section to the test documentation for expected versus current node behaviour. I am unsure if this might not be better to keep in a separate document instead?

In other words, we could update `spec.md` to always contain the expected behaviour (and therefore what we test), and another document for what currently occurs. Thoughts?

I'll update the PR with whatever we decide (I currently haven't updated the spec either).

### DIFF
--- a/src/tests/conformance/handshake.rs
+++ b/src/tests/conformance/handshake.rs
@@ -343,6 +343,81 @@ async fn reject_version_reusing_nonce() {
     node.stop().await;
 }
 
+#[tokio::test]
+async fn reject_obsolete_versions() {
+    // ZG-CONFORMANCE-007
+    //
+    // The node rejects connections with obsolete node versions.
+    //
+    // We expect the following behaviour, regardless of who initiates the connection:
+    //
+    //  1. We send `version` with an obsolete version number
+    //  2. The node responds with `Reject(Obsolete)`
+    //  3. The node terminates the connection
+    //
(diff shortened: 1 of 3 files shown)


src/tests/conformance/handshake.rs
```diff
@@ -343,6 +343,79 @@ async fn reject_version_reusing_nonce() {
...
+    //
+    // The node rejects connections with obsolete node versions.
+    //
+    // We expect the following behaviour, regardless of who initiates the connection:
+    //
+    //  1. We send `version` with an obsolete version number
+    //  2. The node responds with `Reject(Obsolete)`
+    //  3. The node terminates the connection
+    //
+    // Neither Zebra nor ZCashd currently fully comply with this behaviour, so we may need
```
niklaslong, line 358: Could we leave a clear comment that this test is expected to fail in its current state? Or maybe a `TODO: confirm expected behaviour.` at the top just so we have a consistent way of keeping track of these?

niklaslong (commented):
I made a small suggestion but otherwise looks good 👍

src/tests/conformance/handshake.rs
Mirko-von-Leipzig: Updated with additional comments

niklaslong (approved):

PR #17, opened 2021-04-16
Node start signalling

niklaslong:
This PR abstracts node start signalling through a peer connection. Note the address from the `Config` still needs to be passed in at this stage.

### DIFF
--- a/src/setup/node.rs
+++ b/src/setup/node.rs
@@ -79,7 +79,7 @@ impl Node {
 
     /// Sets whether to signal the node has started through a peer connection.
     ///
-    /// If set to true, the call to [`start`] will initiate a listener set as a peer on the node and
+    /// If set the call to [`start`] will initiate a listener set as a peer on the node and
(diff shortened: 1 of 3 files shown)


src/setup/node.rs
```diff
@@ -74,13 +77,36 @@ impl Node {
...
     }
 
+    /// Sets whether to signal the node has started through a peer connection.
+    ///
+    /// If set to true, the call to [`start`] will initiate a listener set as a peer on the node and
+    /// will only return once it has received a connection request. This isn't necessary in
+    /// scenarios in which the node initates the connections.
+    ///
+    /// [`start`]: methode@Node::start
+    pub fn signal_when_started(&mut self, addr: SocketAddr) -> &mut Self {
```
Mirko-von-Leipzig, line 87: I'm trying to think of a name that better describes that we are waiting for an initial connection signal before returning.

`block_until_connected` ? `wait_until_connected` ?

Mirko-von-Leipzig (changes requested):

PR #19, opened 2021-04-19
feat: added conformance test 010

Mirko-von-Leipzig:
This is only partially implemented and currently fails for both node
types.

In particular we get no response to `GetAddr` which makes me think something else is wrong.

Not certain if we should push this in or not.

### DIFF
--- a/src/tests/conformance/messages.rs
+++ b/src/tests/conformance/messages.rs
@@ -184,6 +184,79 @@ async fn ignores_unsolicited_responses() {
     node.stop().await;
 }
 
+#[tokio::test]
+async fn responds_to_queries() {
+    // ZG-CONFORMANCE-010
+    //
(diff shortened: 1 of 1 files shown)


niklaslong:
I think we should hold off on merging until we've investigated the `GetAddr` situation. I'm thinking it might be due to [our use](https://github.com/eqlabs/ziggurat/blob/7432331e0dac284b9eb97be362bf5de8cf1e4c27/src/setup/config.rs#L250) of `connect` in the config ([doc](https://zcash.readthedocs.io/en/latest/rtd_pages/zcash_conf_guide.html))?
It might be worth looking into the other peering tests (12 and 13) as well.

Mirko-von-Leipzig:
Closing pending investigation.

PR #20, opened 2021-04-19
Feat/progress readme

Mirko-von-Leipzig:
You can view the readme here: https://github.com/eqlabs/ziggurat/tree/feat/progress_readme

Note that the badges are linked to the relevant test case in the spec.

Any feedback on ambiguity or missing info etc is welcome.


niklaslong (approved):
LGTM 👍

PR #22, opened 2021-04-19
feat/conformance 005

Mirko-von-Leipzig:
Test currently does not pass:

- zcashd sends messages despite handshake being incomplete
- zebra terminates the connection (I'm expecting a `Reject`, then termination)

Something we need clarity on, but I want to get the base in so we know.

### DIFF
--- a/src/tests/conformance/handshake.rs
+++ b/src/tests/conformance/handshake.rs
@@ -278,6 +278,110 @@ async fn reject_non_version_replies_to_version() {
 }
 
 #[tokio::test]
+async fn reject_non_verack_replies_to_verack() {
+    // Conformance test 005.
+    //
+    // The node rejects non-Verack message as a response to initial Verack it sent.
+    //
+    // Test procedure:
+    //  For each non-verack message,
+    //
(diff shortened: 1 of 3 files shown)


src/tests/conformance/handshake.rs
```diff
@@ -277,6 +277,110 @@ async fn reject_non_version_replies_to_version() {
...
+    //  3. Expect `Verack`
+    //  4. Send test message
+    //  5. Expect `Reject(Invalid)`
+    //  6. Expect connection to be terminated
+    //
+    // This test currently fails as neither Zebra nor ZCashd currently fully comply
+    // with this behaviour, so we may need to revise our expectations.
+    //
+    // ZCashd node eagerly sends messages before handshake has been concluded.
+    // Zebra node does not send Reject, but terminates the connection.
```
niklaslong, line 300: This may not be implemented yet in Zebra—I had a quick peruse and couldn't find any `Reject` messages being sent.

niklaslong (approved):
LGTM 👍

PR #23, opened 2021-04-20
fix: timestamp codec is i64 instead of u32

Mirko-von-Leipzig:
The chrono timestamp is `i64` but the protocol should encode it as `u32`

### DIFF
--- a/src/protocol/payload/addr.rs
+++ b/src/protocol/payload/addr.rs
@@ -1,5 +1,7 @@
 use crate::protocol::payload::{read_n_bytes, VarInt};
 
+use std::convert::TryInto;
+
 use chrono::{DateTime, NaiveDateTime, Utc};
 
 use std::{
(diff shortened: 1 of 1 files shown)


src/protocol/payload/addr.rs
```diff
@@ -82,8 +84,8 @@ impl NetworkAddr {
     }
 
     pub(super) fn decode(bytes: &mut Cursor<&[u8]>) -> io::Result<Self> {
-        let timestamp = i64::from_le_bytes(read_n_bytes(bytes)?);
-        let dt = DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(timestamp, 0), Utc);
+        let timestamp = u32::from_le_bytes(read_n_bytes(bytes)?);
+        let dt = DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(timestamp.into(), 0), Utc);
```
niklaslong, line 88: I think we should drop the `DateTime` here completely, it's an unnecessary conversion we should only implement for debugging if we need a more human friendly format.

niklaslong (commented):
Good catch! Looks good other than a small suggestion.

src/protocol/payload/addr.rs
Mirko-von-Leipzig: We do still need to initialize new addresses with a timestamp value for tests. Which means we will likely use `chrono::Utc::now()` to get a reasonable value (which we have to convert anyway).

I also prefer the field indicating what it is - and the encoding / decoding already encapsulates what to do with the value.

niklaslong (approved):

PR #24, opened 2021-04-20
feat/conformance 012

Mirko-von-Leipzig:


src/tests/conformance/messages.rs
```diff
@@ -184,6 +184,129 @@ async fn ignores_unsolicited_responses() {
...
+    let mut peer_handles = Vec::with_capacity(listeners.len());
+    for listener in listeners {
+        peer_handles.push(tokio::time::timeout(
+            tokio::time::Duration::from_secs(120),
+            tokio::spawn(async move {
+                println!(
+                    "{:?}: waiting for connection",
+                    listener.local_addr().unwrap()
+                );
+                listener.accept().await.unwrap();
```
niklaslong, line 248: We may want to check the handshake completes here? 
```suggestion
                respond_to_handshake(listener).await.unwrap();
```

src/tests/conformance/messages.rs
```diff
@@ -184,6 +184,129 @@ async fn ignores_unsolicited_responses() {
...
+        let message = filter.read_from_stream(&mut stream).await.unwrap();
+        panic!(
+            "Unexpected message received by main connection: {:?}",
+            message
+        );
+    });
+
+    // wait for peer connections to complete, or main connection to break
+    tokio::select! {
+        _ = main_responder => {},
```
niklaslong, line 303: I don't think this will propagate the panic if the task panics, meaning the test would pass either way? 
Assuming the task only ever panics if it receives an unexpected message, we could simply do:
```suggestion
        _ = main_responder => panic!("..."),
```

src/tests/conformance/messages.rs
```diff
@@ -184,6 +184,129 @@ async fn ignores_unsolicited_responses() {
...
+            handle.await.unwrap().unwrap();
+        }
+    });
+
+    // We need to keep responding to ping requests on the main connection,
+    // otherwise it may get marked as unreliable (and the peer list gets ignored).
+    //
+    // Without this, zebra takes forever to connect and spuriously fails as well.
+    // TBC - this is all speculation.
+    let main_responder = tokio::spawn(async move {
```
niklaslong, line 290: Great! This made a huge difference for me locally: test runs in under one second.

src/tests/conformance/messages.rs
```diff
@@ -184,6 +184,129 @@ async fn ignores_unsolicited_responses() {
...
+    //  3. Wait for `GetAddr`
+    //  4. Send set of peer listener node addresses
+    //  5. Expect the node to connect to each peer in the set
+    //
+    // This test currently fails for zcashd; zebra passes (with a caveat).
+    //
+    // Current behaviour:
+    //
+    //  zcashd: Has different behaviour depending on connection direction.
+    //          If we initiate the main conneciton it sends Ping, GetHeaders,
```
niklaslong, line 206:
```suggestion
    //          If we initiate the main connection it sends Ping, GetHeaders,
```

niklaslong (changes requested):
Great work on this! It's also fast now 🚀 
I left a few suggestions, the main thing is we need to catch the panic if the main responder breaks.

src/tests/conformance/messages.rs
Mirko-von-Leipzig: Sure we can do that.

niklaslong (approved):

PR #25, opened 2021-04-23
feat/conformance 013

Mirko-von-Leipzig:
I know we said not to add many peers, but I ended up writing it as a loop (was shorter codewise) so simply a matter of setting N.


niklaslong (changes requested):
This test currently doesn't complete locally for me against zebra? It just seems to crawl and dial forever?

niklaslong (approved):

PR #27, opened 2021-04-29
feat: added block hashing

Mirko-von-Leipzig:
I'm not totally stoked about the naming of the hash function - `double_hash`.

### DIFF
--- a/src/protocol/payload/block.rs
+++ b/src/protocol/payload/block.rs
@@ -1,6 +1,11 @@
 use crate::protocol::payload::{read_n_bytes, Hash, ProtocolVersion, Tx, VarInt};
 
-use std::io::{self, Cursor, Write};
+use std::{
+    convert::TryInto,
+    io::{self, Cursor, Write},
+};
+
+use sha2::Digest;
 
 #[derive(Debug)]
 pub struct LocatorHashes {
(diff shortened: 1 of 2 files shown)


src/protocol/payload/mod.rs
```diff
@@ -126,10 +126,14 @@ impl VarStr {
     }
 }
 
-#[derive(Debug, PartialEq)]
-struct Hash([u8; 32]);
+#[derive(Debug, PartialEq, Clone)]
```
niklaslong, line 129: Is `Clone` here necessary?

src/protocol/payload/block.rs
```diff
@@ -176,10 +174,42 @@ impl Header {
...
+        buffer.write_all(&self.nonce)?;
+
+        self.solution_size.encode(buffer)?;
+        buffer.write_all(&self.solution)?;
+
+        Ok(())
+    }
+
+    /// Calculates the double Sha256 hash for [Header]
+    fn double_hash(&self) -> std::io::Result<Hash> {
```
niklaslong, line 195: What about `double_sha256`?

src/protocol/payload/block.rs
```diff
@@ -2,6 +2,10 @@ use crate::protocol::payload::{read_n_bytes, Hash, ProtocolVersion, Tx, VarInt};
 
 use std::io::{self, Cursor, Write};
 
+use sha2::Digest;
+
+use std::convert::TryInto;
```
niklaslong, line 7: I'd consolidate this with the `use std...` above ☝️

niklaslong (commented):
LGTM with a few comments 👍

src/protocol/payload/mod.rs
Mirko-von-Leipzig: Ah its not currently. I had use for it temporarily, will remove.

niklaslong (approved):

PR #29, opened 2021-05-03
test: resistance-001/002

niklaslong:
This PR introduces the first fuzz test cases. 

Included: 
- resistance-001 for all types of payloads,
- resistance-002 for all types of payloads,
- adjusts `Message` and `MessageHeader` to allow encoding/decoding without writing to stream.
- introduces various helpers for fuzzing payload creation

### DIFF
--- a/src/tests/resistance/fuzzing.rs
+++ b/src/tests/resistance/fuzzing.rs
@@ -15,15 +15,8 @@ use crate::{
     setup::{config::read_config_file, node::Node},
 };
 
-use tokio::{
-    io::{self, AsyncWriteExt},
-    net::TcpStream,
-    time::timeout,
-};
-
 use rand::{distributions::Standard, prelude::SliceRandom, thread_rng, Rng};
-
-use std::{net::SocketAddr, time::Duration};
+use tokio::{io::AsyncWriteExt, net::TcpStream};
 
 const ITERATIONS: usize = 100;
 const CORRUPTION_PROBABILITY: f64 = 0.5;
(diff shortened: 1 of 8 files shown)


src/tests/resistance/fuzzing.rs
```diff
@@ -0,0 +1,813 @@
...
+use crate::{
+    helpers::{autorespond_and_expect_disconnect, initiate_version_exchange},
+    protocol::{
+        message::*,
+        payload::{block::Headers, Addr, Nonce, Version},
+    },
+    setup::{config::read_config_file, node::Node},
+};
+
+use rand::{distributions::Standard, prelude::SliceRandom, thread_rng, Rng};
```
Mirko-von-Leipzig, line 18: It may be necessary to store the RNG seed for each trial - and/or the message contents required to reproduce the failure.

Since each run will be unique (I think?), if something does fail at some stage we will want to know what the seed for that instance was.

Mirko-von-Leipzig (changes requested):
LGTM, nice work! If we can just discuss the reproducibility aspect.

Mirko-von-Leipzig (changes requested):
LGTM, nice work! If we can just discuss the reproducibility aspect.

src/tests/resistance/fuzzing.rs
niklaslong: Good point, I'll look into this 👍

Mirko-von-Leipzig (approved):

PR #31, opened 2021-05-05
fix: Header tx_count must be zero for encoding

Mirko-von-Leipzig:
The spec states that the `tx_count` is always zero when transmitted in the Header message.

This removes the `tx_count` from the header field, instead opting to encode / decode the VarInt according to its purpose at the next highest level.

So for a `Block` message it will encode the actual tx count, and for `Headers` each `Header` will have a `VarInt(0)` appended.

### DIFF
--- a/src/protocol/payload/addr.rs
+++ b/src/protocol/payload/addr.rs
@@ -1,4 +1,4 @@
-use crate::protocol::payload::{read_n_bytes, VarInt};
+use crate::protocol::payload::{codec::Codec, read_n_bytes};
 
 use std::convert::TryInto;
 
@@ -23,31 +23,19 @@ impl Addr {
         Addr { addrs }
     }
 
-    pub fn encode(&self, buffer: &mut Vec<u8>) -> io::Result<()> {
-        VarInt(self.addrs.len()).encode(buffer)?;
-
-        for addr in &self.addrs {
-            addr.encode(buffer)?;
-        }
-
-        Ok(())
-    }
-
-    pub fn decode(bytes: &mut Cursor<&[u8]>) -> io::Result<Self> {
-        let count = *VarInt::decode(bytes)?;
-        let mut addrs = Vec::with_capacity(count);
-
-        for _ in 0..count {
-            let addr = NetworkAddr::decode(bytes)?;
-            addrs.push(addr);
-        }
-
-        Ok(Self::new(addrs))
-    }
-
     /// Returns an iterator over its list of [NetworkAddr]'s
     pub fn iter(&self) -> std::slice::Iter<NetworkAddr> {
         self.addrs.iter()
+    }
+}
+
+impl Codec for Addr {
+    fn encode(&self, buffer: &mut Vec<u8>) -> io::Result<()> {
+        self.addrs.encode(buffer)
+    }
+
+    fn decode(bytes: &mut Cursor<&[u8]>) -> io::Result<Self> {
+        Ok(Self::new(Vec::decode(bytes)?))
     }
 }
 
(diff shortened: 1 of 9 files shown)


Mirko-von-Leipzig:
I'm tempted to remove all `VarInt` fields, since the length is always implicitly stored already by the vector. However, I wondered if we don't want to keep it for validation purposes? Although in that case the decoding and/or encoding step would have failed.

Basically its redundant information that I want to strip out, and move into the encode / decode methods.

src/protocol/payload/codec.rs
```diff
@@ -0,0 +1,29 @@
+use super::VarInt;
+
+pub trait Codec {
+    fn encode(&self, buffer: &mut Vec<u8>) -> std::io::Result<()>;
+    fn decode(bytes: &mut std::io::Cursor<&[u8]>) -> std::io::Result<Self>
+    where
+        Self: Sized;
+}
+
+impl<T: Codec> Codec for Vec<T> {
```
Mirko-von-Leipzig, line 10: This `impl` was the main motivation for creating the Trait now.

src/protocol/payload/mod.rs
```diff
@@ -74,33 +80,29 @@ impl std::ops::Deref for VarInt {
     }
 }
 
-impl VarInt {
-    fn encode(&self, buffer: &mut Vec<u8>) -> io::Result<usize> {
```
Mirko-von-Leipzig, line 78: Note the removal of `<usize>` -- required to meet the Trait definitation, and I couldn't find a use case for this (?).

src/protocol/payload/mod.rs
```diff
@@ -121,11 +123,11 @@ impl VarInt {
 struct VarStr(String);
 
 impl VarStr {
-    fn encode(&self, buffer: &mut Vec<u8>) -> io::Result<usize> {
-        let int_len = VarInt(self.0.len()).encode(buffer)?;
```
Mirko-von-Leipzig, line 125: Note the removal of `<usize>` -- required to meet the Trait definitation, and I couldn't find a use case for this (?).

Mirko-von-Leipzig:
There are two rough spots remaining

1. Headers can't use the Vec Trait imple because of the extra VarInt(0) we have to append.
2. Vec<JoinSplit> needs JoinSplit to not have different encode_xxx methods

src/protocol/payload/addr.rs
```diff
@@ -9,50 +9,35 @@ use std::{
     net::{IpAddr::*, Ipv6Addr, SocketAddr},
 };
 
+use super::codec::Codec;
```
niklaslong, line 12: In general, I'd consolidate the imports using absolute paths as we're already importing `protocol::payload::{...}`

src/protocol/payload/block.rs
```diff
@@ -7,10 +7,11 @@ use std::{
 
 use sha2::Digest;
 
+use super::codec::Codec;
```
niklaslong, line 10: Same as above regarding imports, there are a few other spots as well.

src/protocol/payload/mod.rs
niklaslong: Likely left over from calculating the message length for the header, it's not needed anymore 👌

niklaslong (approved):
LGTM 🎉 
`JoinSplit` needs the `decode` both for groth16 and bctv14, this is linked to the tx version, hence the need for two functions accessible on the tx level. The inability to implement it with the `Codec` trait is what bothered me at the time. Not sure what the best solution is here. Regarding the headers, I think it's reasonable to encode the `VarInt` manually as it is a constant value.

src/protocol/payload/addr.rs
Mirko-von-Leipzig: Good point, I just used RA's auto-import 💡 which did that. Fixed.

PR #33, opened 2021-05-06
ref: seedable RNG for fuzz test reproducability

niklaslong:
Follow up to #29. Replaces the RNG with the seedable `ChaCha8Rng`. This makes the tests reproducable.

### DIFF
--- a/src/tests/resistance/fuzzing.rs
+++ b/src/tests/resistance/fuzzing.rs
@@ -15,7 +15,12 @@ use crate::{
     setup::{config::read_config_file, node::Node},
 };
 
-use rand::{distributions::Standard, prelude::SliceRandom, thread_rng, Rng};
+use rand::{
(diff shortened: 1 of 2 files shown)


src/tests/resistance/fuzzing.rs
```diff
@@ -5,6 +5,10 @@
 // - Slightly corrupted but otherwise valid messages, e.g. N% of body replaced with random bytes.
 // - Messages with an incorrect checksum.
 // - Messages with differing announced and actual lengths.
+//
+// A note on RNGs: each test makes the use of one seeded ChaCha8Rng. If the test fails, the test
```
Mirko-von-Leipzig, line 9: Small typo `makes the use of one` should just be `makes use of one`

Mirko-von-Leipzig (approved):
LGTM, just the minor typo.

PR #34, opened 2021-05-06
feat/node cleanup

Mirko-von-Leipzig:
I'm expecting this to not be working for zcashd docker -- not sure how to construct the relevant paths there. If you could test and let me know?


Mirko-von-Leipzig:
We may also want to selectively delete zcash files. For now I'm just nuking all of it.

niklaslong (approved):
Nuking the cache is reasonable, especially as we want clean nodes for each test. 
We should drop docker support, so this isn't a worry either.

PR #35, opened 2021-05-06
test: resistance-005

niklaslong:
Some cases are tbd as they require more in-depth investigations.

### DIFF
--- a/src/tests/resistance/fuzzing.rs
+++ b/src/tests/resistance/fuzzing.rs
@@ -302,7 +302,8 @@ async fn fuzzing_metadata_compliant_random_bytes_post_handshake() {
     // ZG-RESISTANCE-005 (part 3)
     //
     // zebra: breaks with a version command in header, spams getdata, doesn't disconnect.
-    // zcashd:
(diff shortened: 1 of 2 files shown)


src/tests/resistance/fuzzing.rs
```diff
@@ -398,6 +539,50 @@ async fn fuzzing_slightly_corrupted_messages_during_handshake_responder_side() {
...
+    // zebra: sends getdata and ignores message.
+    // zcashd: tbd.
+
+    let test_messages = vec![
+        Message::MemPool,
+        Message::Verack,
+        Message::Ping(Nonce::default()),
+        Message::Pong(Nonce::default()),
+        Message::GetAddr,
+        Message::Addr(Addr::empty()),
```
Mirko-von-Leipzig, line 555: I wonder if we shouldn't add variants with non-empty contents as well (where an element is corrupted).

Mirko-von-Leipzig (approved):
LGTM! Nice job.

src/tests/resistance/fuzzing.rs
niklaslong: Yes, good idea, will add as a follow up 👍

PR #36, opened 2021-05-07
Add action to seed node with genesis blocks on startup

Mirko-von-Leipzig:
Refactors the `node.wait_for_connection` to `node.initial_action(Action)` where `Action` can be one of

1. None
2. Wait for connection
3. Seed with genesis blocks

One alternative is to create a helper function instead. I decided it would be better to have it as part of the node startup since the procedure requires that no other data has been sent to the node yet.

I'm open to changing this though.

### DIFF
--- a/src/setup/config.rs
+++ b/src/setup/config.rs
@@ -84,7 +84,7 @@ impl NodeConfig {
 }
 
 /// Describes the node kind, currently supports the two known variants.
-#[derive(Deserialize, Debug, Clone, Copy)]
+#[derive(Deserialize, Debug, Clone, Copy, PartialEq)]
 #[serde(rename_all(deserialize = "lowercase"))]
 pub(super) enum NodeKind {
     Zebra,
--- a/src/setup/config.rs
+++ b/src/setup/config.rs
@@ -9,6 +9,8 @@ use std::{
     path::PathBuf,
 };
 
+use crate::setup::node::Action;
+
 const NODE_PORT: u16 = 8080;
 
 /// Reads the contents of Ziggurat's configuration file.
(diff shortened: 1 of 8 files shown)


src/setup/node.rs
```diff
@@ -131,9 +152,89 @@ impl Node {
...
+                    Message::GetData(inv) => {
+                        // The request must be for blocks [1,2]
+                        let expected = Inv::new(vec![
+                            InvHash::new(ObjectKind::Block, block_1.double_sha256().unwrap()),
+                            InvHash::new(ObjectKind::Block, block_2.double_sha256().unwrap()),
+                        ]);
+
+                        assert_eq!(inv, expected);
+
+                        // send blocks 1 and 2
```
niklaslong, line 214: Should the genesis block be sent as well? I'm not sure nodes come pre-seeded with the genesis block?

niklaslong (commented):
Left a few comments, otherwise this is looking great 🎉

src/setup/config.rs
```diff
@@ -9,6 +9,8 @@ use std::{
     path::PathBuf,
 };
 
+use super::node::Action;
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I blame RA :D

src/setup/node.rs
Mirko-von-Leipzig: I was wondering about that, but the node never actually requests it? Initially I expected `inv` to contain a request for all three blocks, but it only requests the last two.

niklaslong: Per discussion, seems like it is ignored by the nodes when sent: Nodes must be pre-seeded with the block.

PR #37, opened 2021-05-07
feat: added conformance test 016

Mirko-von-Leipzig:


src/tests/conformance/messages.rs
```diff
@@ -372,6 +377,125 @@ async fn correctly_lists_peers() {
...
+    for (start, stop) in ranges {
+        Message::GetHeaders(LocatorHashes::new(
+            locator[start].clone(),
+            hashes[stop].clone(),
+        ))
+        .write_to_stream(&mut stream)
+        .await
+        .unwrap();
+
+        // requried to handle cases where start and stop are the same
```
niklaslong, line 467: Could you extend the comment to include why there's a need to shift the expected `start_i` index checked in the response if the `start` and `stop` indices are the same?

niklaslong (commented):
LGTM, with a few comments 👍

src/tests/conformance/messages.rs
Mirko-von-Leipzig: Looking at this with a fresh(er) mind, I no longer think its required at all.

niklaslong (approved):

PR #39, opened 2021-05-11
conformance 015

Mirko-von-Leipzig:


src/protocol/payload/block.rs
```diff
@@ -174,17 +183,15 @@ impl Codec for Header {
 
 impl Header {
     /// Calculates the double Sha256 hash for [Header]
-    pub fn double_sha256(&self) -> std::io::Result<Hash> {
+    pub fn double_sha256(&self) -> Hash {
         let mut buffer = Vec::new();
 
-        self.encode_without_tx_count(&mut buffer)?;
+        self.encode_without_tx_count(&mut buffer).unwrap();
```
niklaslong, line 189: I'm not convinced we'd want to unwrap this? In general having protocol implementation code return errors would be better, we can always unwrap at the test level.

src/protocol/payload/tx.rs
```diff
@@ -21,17 +21,15 @@ pub enum Tx {
 }
 
 impl Tx {
-    pub fn double_sha256(&self) -> std::io::Result<Hash> {
+    pub fn double_sha256(&self) -> Hash {
         let mut buffer = Vec::new();
 
-        self.encode(&mut buffer)?;
+        self.encode(&mut buffer).unwrap();
```
niklaslong, line 27: Same unwrap considerations as above.

src/tests/conformance/messages.rs
```diff
@@ -378,6 +378,116 @@ async fn correctly_lists_peers() {
...
+    //  3. For each test case:
+    //      a) send GetBlocks
+    //      b) receive Inv
+    //      c) assert Inv received matches expectations
+    //
+    // The test currently fails for both Zebra and zcashd.
+    //
+    // Current behaviour:
+    //
+    //  zcashd: Fails because it ignores requests when the last hash is the last hash it knows of. We expect it to return
```
niklaslong, line 404: Wouldn't this be expected with `start < range < stop` as opposed to `<= stop`?

src/tests/conformance/messages.rs
```diff
@@ -378,6 +378,116 @@ async fn correctly_lists_peers() {
...
+    .log_to_stdout(true)
+    .start()
+    .await;
+
+    let blocks = Block::initial_testnet_blocks();
+
+    let mut stream = initiate_handshake(node.addr()).await.unwrap();
+    let filter = MessageFilter::with_all_auto_reply();
+
+    // Test GetBlocks(i) -> Inv(i+1..)
```
niklaslong, line 429: This makes sense to an informed reader but it would be good to have some more context in plain english?

src/tests/conformance/messages.rs
```diff
@@ -378,6 +378,116 @@ async fn correctly_lists_peers() {
...
+            vec![block.double_sha256()],
+            Hash::zeroed(),
+        ))
+        .write_to_stream(&mut stream)
+        .await
+        .unwrap();
+
+        match filter.read_from_stream(&mut stream).await.unwrap() {
+            Message::Inv(inv) => {
+                let inv_hashes = blocks.iter().skip(i + 1).map(|b| b.inv_hash()).collect();
```
niklaslong, line 441: In conjunction with the above, the `i + 1` detail can easily be overlooked.

src/tests/conformance/messages.rs
```diff
@@ -378,6 +378,116 @@ async fn correctly_lists_peers() {
...
+        Message::Inv(inv) => {
+            let expected = Inv::new(vec![blocks[1].inv_hash()]);
+            assert_eq!(inv, expected);
+        }
+        message => panic!("Expected Inv, but got {:?}", message),
+    }
+
+    // Test that we get corrected if we are "off chain".
+    // We expect that unknown hashes get ignored, until it finds a known hash; it then returns
+    // all known children of that block.
```
niklaslong, line 467: Isn't it also supposed to send a `NotFound` in conjunction with the `Inv`?

niklaslong (changes requested):
Great to have this implemented 🎉 
I left a few comments, most notably about limiting the use of `unwrap` in setup and protocol implementation code.

src/tests/conformance/messages.rs
Mirko-von-Leipzig: I would still expect a reply; even if its just the empty set `Inv({})`. 

From the spec:
"If a block hash is found in the node's main chain, the list of its children is returned back via the inv message" - in this case the list is simply empty.

src/tests/conformance/messages.rs
Mirko-von-Leipzig: Good point

src/protocol/payload/block.rs
Mirko-von-Leipzig: Reverted

PR #40, opened 2021-05-11
ref: remove docker configuration support

niklaslong:
This PR: 
- removes docker configuration (see readme for new format),
- removes `read_config_file` (absracted),
- changes `initial_peers` to accept full addresses instead of just the port,
- exposes `config::new_local_addr` as a convenience function,
- introduces constants for configuration file names.

### DIFF
--- a/src/setup/config.rs
+++ b/src/setup/config.rs
@@ -105,8 +105,6 @@ impl NodeMetaData {
     }
 }
 
-// ZEBRA CONFIG FILE
-
 /// Convenience struct for writing a zebra compatible configuration file.
 #[derive(Serialize)]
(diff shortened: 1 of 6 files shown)


src/setup/config.rs
```diff
@@ -237,6 +166,8 @@ struct TracingConfig {
     filter: Option<String>,
 }
 
+// ZCASHD CONFIG FILE
```
Mirko-von-Leipzig, line 169: Maybe remove? Its literally the name of the struct :D

Mirko-von-Leipzig (changes requested):
Nice work! Its a lot cleaner - me likey.

Just the one comment to remove.

PR #45, opened 2021-05-14
conformance test 010 and 009

Mirko-von-Leipzig:
Note that this includes the missing test messages for case 009.

### DIFF
--- a/src/tests/conformance/messages.rs
+++ b/src/tests/conformance/messages.rs
@@ -187,7 +187,7 @@ async fn ignores_unsolicited_responses() {
 
 #[tokio::test]
 async fn basic_query_response() {
-    // ZG-CONFORMANCE-010
+    // ZG-CONFORMANCE-010, node is seeded with data
     //
     // The node responds with the correct messages. Message correctness is naively verified through successful encoding/decoding.
     //
@@ -199,7 +199,7 @@ async fn basic_query_response() {
     // `GetData(block_hash)` expects `Blocks`.
     // `GetHeaders` expects `Headers`.
     //
-    // The test currently fails for zcashd; zebra tbd.
+    // The test currently fails for zcashd and zebra.
     //
     // Current behaviour:
     //
@@ -207,18 +207,19 @@ async fn basic_query_response() {
     //              - GetAddr
     //              - MemPool
     //              - GetBlocks
-    //              - GetData(block)
     //
     //          GetData(tx) returns NotFound (which is correct),
     //          because we currently can't seed a mempool.
     //
-    //  zebra:  tbd
+    //  zebra: DDoS spam due to auto-response
 
     let mut node: Node = Default::default();
-    node.initial_action(Action::WaitForConnection(new_local_addr()))
-        .log_to_stdout(true)
-        .start()
-        .await;
+    node.initial_action(Action::SeedWithTestnetBlocks {
+        socket_addr: new_local_addr(),
+        block_count: 3,
+    })
+    .start()
+    .await;
 
     let mut stream = initiate_handshake(node.addr()).await.unwrap();
     let filter = MessageFilter::with_all_auto_reply();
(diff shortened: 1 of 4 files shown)


src/protocol/message.rs
```diff
@@ -104,7 +104,48 @@ pub enum Message {
     Reject(Reject),
 }
 
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum MessageKind {
```
Mirko-von-Leipzig, line 108: I'm on the fence about this -- I know its not required, but it allows changing the following `assert` style:

Before:
```
assert!(matches!(message, Message::Block(..)), "Expected Block, got {:?}", message);
```
where the formatting is required to get a reasonable panic message.

After:
```
assert_eq!(message.kind(), MessageKind::Block);
```

src/tests/conformance/messages.rs
```diff
@@ -628,7 +714,7 @@ async fn get_data_blocks() {
     let mut node: Node = Default::default();
     node.initial_action(Action::SeedWithTestnetBlocks {
         socket_addr: new_local_addr(),
-        block_count: 3,
+        block_count: 2,
```
niklaslong, line 717: Just making sure this isn't an accidental change?

src/tests/conformance/messages.rs
```diff
@@ -178,15 +177,102 @@ async fn ignores_unsolicited_responses() {
...
     }
 
     node.stop().await;
 }
 
+#[tokio::test]
+async fn basic_query_response() {
+    // ZG-CONFORMANCE-010
+    //
+    // The node responds with the correct messages. Message correctness is naively verified through successful encoding/decoding.
```
niklaslong, line 191: This test is mostly covered by more specific cases but I think it has value in testing explicit expectations around how a node should respond to queries (at least while the spec isn't clearly defined)?

niklaslong (commented):
Looks great with a few comments, the main thing being the message `kind` change.

src/protocol/message.rs
Mirko-von-Leipzig: We still need to match in some instances (to check the message internals). Simple current example:
```
match message {
   Message::Pong(nonce) => assert_eq!(nonce, expected_nonce),
   msg => panic!("Expected Pong, got {:?}, msg),
}
```
So we want something where we can easily destructure into content as well as assert on the basic type.

I'm not sure there's a way to do both with just fields - I think one will always need an additional method.

src/tests/conformance/messages.rs
```diff
@@ -178,15 +177,102 @@ async fn ignores_unsolicited_responses() {
...
+    // `Getblocks` expects `Inv`.
+    // `GetData(tx_hash)` expects `Tx`.
+    // `GetData(block_hash)` expects `Blocks`.
+    // `GetHeaders` expects `Headers`.
+    //
+    // The test currently fails for zcashd; zebra tbd.
+    //
+    // Current behaviour:
+    //
+    //  zcashd: Ignores the following messages
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: Its.. inconsistent. I did initially seed it with blocks, and then `GetBlocks`, and `GetData(block)` work as expected.

Then one has `GetData(tx)` which actually returns `NotFound` despite not having data.

I would actually hope that all of these queries should result in an `{empty}` style response, or `NotFound` for the `GetData`.

Essentially I think the node never sends an empty response except for `GetData(tx)` - because the bitcoin protocol lists it as an example:
```
notfound is a response to a getdata, sent if any requested data items could not be relayed, for example, because the requested transaction was not in the memory pool or relay set. 
```

src/tests/conformance/messages.rs
Mirko-von-Leipzig: It is certainly redundant in terms of coverage. But I like having it as it as it highlights discrepancies between responses.

I think we may want to refactor our test spec and contents once we have the spec more nailed down.

src/tests/conformance/messages.rs
Mirko-von-Leipzig: Good catch.

src/tests/conformance/messages.rs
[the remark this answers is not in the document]
niklaslong: Might be good to have both cases? This would also serve as documentation for the current behaviour in each case and make it easier for ECC/ZF to evaluate what the spec should be.

src/tests/conformance/messages.rs
niklaslong: Agreed 👍

Mirko-von-Leipzig:
I've changed to using a `macro` for the enum asserts. Let me know what you think.

src/tests/mod.rs
```diff
@@ -1,2 +1,22 @@
...
+///
+/// Useful for asserting a enum matches a specific variant.
+///
+/// Usage:
+///     `assert_matches!(value, MyEnum::Variant(..));`
+///
+/// or with additional context:
+///     `assert_matches!(value, MyEnum::Variant(..), "Additional {}", "context");`
+#[macro_export]
+macro_rules! assert_matches {
```
niklaslong, line 14: Lovely! Here's the [std implementation](https://doc.rust-lang.org/nightly/src/core/macros/mod.rs.html#140-165) and [tracking issue](https://github.com/rust-lang/rust/issues/82775) of this if you're curious.

src/tests/conformance/messages.rs
```diff
@@ -178,15 +178,102 @@ async fn ignores_unsolicited_responses() {
...
+    //              - GetData(block)
+    //
+    //          GetData(tx) returns NotFound (which is correct),
+    //          because we currently can't seed a mempool.
+    //
+    //  zebra:  tbd
+
+    let mut node: Node = Default::default();
+    node.initial_action(Action::WaitForConnection(new_local_addr()))
+        .log_to_stdout(true)
```
niklaslong, line 219:
```suggestion
```

src/tests/conformance/messages.rs
```diff
@@ -178,15 +178,102 @@ async fn ignores_unsolicited_responses() {
...
+    Message::GetHeaders(LocatorHashes::new(
+        vec![genesis_block.double_sha256().unwrap()],
+        Hash::zeroed(),
+    ))
+    .write_to_stream(&mut stream)
+    .await
+    .unwrap();
+    let reply = filter.read_from_stream(&mut stream).await.unwrap();
+    assert_matches!(reply, Message::Headers(..));
+}
```
niklaslong, line 275:
```suggestion
node.stop().await;
}
```

niklaslong (changes requested):
Looking good other than a few small things, love the macro 🎉

src/tests/conformance/messages.rs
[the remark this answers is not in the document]
Mirko-von-Leipzig: I'll split up the test case into seeded and unseeded

niklaslong (approved):

PR #49, opened 2021-05-18
doc: fuzz test comments

niklaslong:
Follow up to #35.


Mirko-von-Leipzig (approved):
Nice work. Would be nice to tell which messages cause what. Falls in the nice-to-have test runner bin (:

PR #50, opened 2021-05-19
Added resistance test cases 003 and 004

Mirko-von-Leipzig:
I still need to implement the missing message types. Thought I'd first get it on par with the rest before we add / refactor the rest of the message types.

### DIFF
--- a/src/tests/resistance/fuzzing.rs
+++ b/src/tests/resistance/fuzzing.rs
@@ -1792,8 +1792,155 @@ async fn fuzzing_version_with_incorrect_length_during_handshake_responder_side()
 }
 
 #[tokio::test]
+async fn fuzzing_version_with_incorrect_length_when_node_initiates_handshake() {
+    // ZG-RESISTANCE-003 (part 6, version only)
+    //
+    // This particular case is considered alone because it is at particular risk of causing
+    // troublesome behaviour, as seen with the valid metadata fuzzing against zebra.
+    //
+    // zebra: disconnects immediately.
+    // zcashd: Some messages get ignored and timeout. Most cause an immedietely due to
+    //          - main: PROCESSMESSAGE: INVALID MESSAGESTART, or
+    //          - net: Oversized message from peer
+
+    let locked_rng = Arc::new(Mutex::new(seeded_rng()));
+
+    // create tcp listeners for peer set (port is only assigned on tcp bind)
+    let mut listeners = Vec::with_capacity(ITERATIONS);
+    for _ in 0..ITERATIONS {
+        listeners.push(TcpListener::bind(new_local_addr()).await.unwrap());
+    }
+
+    // get list of peer addresses to pass to node
+    let peer_addresses = listeners
+        .iter()
+        .map(|listener| listener.local_addr().unwrap())
+        .collect::<Vec<_>>();
+
+    // start peer processes
+    let mut peer_handles = Vec::with_capacity(listeners.len());
+    for peer in listeners {
+        let peer_rng = locked_rng.clone();
+        peer_handles.push(tokio::time::timeout(
+            tokio::time::Duration::from_secs(120),
+            tokio::spawn(async move {
+                // Await connection and receive version
+                let (mut peer_stream, _) = peer.accept().await.unwrap();
+                let version = Message::read_from_stream(&mut peer_stream).await.unwrap();
+                assert_matches!(version, Message::Version(..));
+
+                // send bad version
+                let mut rng = peer_rng.lock_owned().await;
+                let version = Message::Version(Version::try_from_stream(&peer_stream).unwrap());
+                let mut message_buffer = vec![];
+                let mut header = version.encode(&mut message_buffer).unwrap();
+
+                // Set the length to a random value which isn't the current value.
+                header.body_length = random_non_valid_u32(&mut rng, header.body_length);
+
+                let _ = header.write_to_stream(&mut peer_stream).await;
+                let _ = peer_stream.write_all(&message_buffer).await;
+
+                autorespond_and_expect_disconnect(&mut peer_stream).await;
+            }),
+        ));
+    }
+
+    let mut node: Node = Default::default();
+    node.initial_action(Action::None)
+        .initial_peers(peer_addresses)
+        .start()
+        .await;
+
+    // join the peer processes
+    for handle in peer_handles {
+        handle.await.unwrap().unwrap();
+    }
+
+    node.stop().await;
+}
+
+#[tokio::test]
+async fn fuzzing_version_with_incorrect_length_inplace_of_verack_when_node_initiates_handshake() {
+    // ZG-RESISTANCE-004 (part 6, version only)
+    //
+    // This particular case is considered alone because it is at particular risk of causing
+    // troublesome behaviour, as seen with the valid metadata fuzzing against zebra.
+    //
+    // zebra: disconnects immediately.
+    // zcashd: Sends GetAddr, Ping and GetHeaders. Appears to ignore bad verack message.
+
+    let locked_rng = Arc::new(Mutex::new(seeded_rng()));
+
+    // create tcp listeners for peer set (port is only assigned on tcp bind)
+    let mut listeners = Vec::with_capacity(ITERATIONS);
+    for _ in 0..ITERATIONS {
+        listeners.push(TcpListener::bind(new_local_addr()).await.unwrap());
+    }
+
+    // get list of peer addresses to pass to node
+    let peer_addresses = listeners
+        .iter()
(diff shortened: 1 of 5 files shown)


src/tests/resistance/fuzzing.rs
```diff
@@ -398,6 +820,143 @@ async fn fuzzing_slightly_corrupted_version_during_handshake_responder_side() {
...
+    //
+    // This particular case is considered alone because it is at particular risk of causing
+    // troublesome behaviour, as seen with the valid metadata fuzzing against zebra.
+    //
+    // zebra: disconnects immediately.
+    // zcashd: Some messages get ignored and timeout. Most cause an immedietely due to
+    //          - main: PROCESSMESSAGE: INVALID MESSAGESTART, or
+    //          - net: Oversized message from peer
+
+    let locked_rng = Arc::new(Mutex::new(seeded_rng()));
```
Mirko-von-Leipzig, line 835: I'm not sure there's a different way of doing this?

The issue is that I can't generate the payloads ahead of time (since the payload relies on both the node and peer addresses). This means `rng` is needed inside each future / concurrently.

Overall not an issue, just something I'm happy to know if I did correctly.

src/tests/resistance/fuzzing.rs
```diff
@@ -85,6 +91,132 @@ async fn fuzzing_zeroes_during_handshake_responder_side() {
...
+        peer_handles.push(tokio::time::timeout(
+            tokio::time::Duration::from_secs(120),
+            tokio::spawn(async move {
+                // Await connection and receive version
+                let (mut peer_stream, _) = peer.accept().await.unwrap();
+                let version = Message::read_from_stream(&mut peer_stream).await.unwrap();
+                assert_matches!(version, Message::Version(..));
+
+                // send version, receive verack
+                Message::Version(Version::new(
```
niklaslong, line 189: `Version::try_from_stream`? Though I quite like the explicitness of constructing the version like this as the function name is a little confusing?

src/tests/resistance/fuzzing.rs
niklaslong: I don't see many solutions, this might be the best with the current setup.

src/tests/resistance/fuzzing.rs
```diff
@@ -21,14 +22,19 @@ use crate::{
...
     distributions::Standard,
     prelude::{Rng, SeedableRng, SliceRandom},
     thread_rng,
 };
 use rand_chacha::ChaCha8Rng;
-use tokio::{io::AsyncWriteExt, net::TcpStream};
+use tokio::{
+    io::AsyncWriteExt,
+    net::{TcpListener, TcpStream},
+    sync::Mutex,
```
niklaslong, line 36: I'd use the parking lot variant of `Mutex` as we don't need to hold the lock over an `await`. I'd also use an `RwLock` instead since the rng is `Sync` + `Send`. https://docs.rs/lock_api/0.4.4/lock_api/struct.Mutex.html

src/tests/resistance/fuzzing.rs
```diff
@@ -398,6 +820,143 @@ async fn fuzzing_slightly_corrupted_version_during_handshake_responder_side() {
...
+        peer_handles.push(tokio::time::timeout(
+            tokio::time::Duration::from_secs(120),
+            tokio::spawn(async move {
+                // Await connection and receive version
+                let (mut peer_stream, _) = peer.accept().await.unwrap();
+                let version = Message::read_from_stream(&mut peer_stream).await.unwrap();
+                assert_matches!(version, Message::Version(..));
+
+                // send bad version
+                let mut rng = peer_rng.lock_owned().await;
```
niklaslong, line 862: Make sure you drop the lock as soon as it isn't needed anymore.

src/tests/resistance/fuzzing.rs
```diff
@@ -1064,7 +2247,6 @@ fn zeroes(rng: &mut ChaCha8Rng, n: usize) -> Vec<Vec<u8>> {
     (0..n)
         .map(|_| {
             let random_len: usize = rng.gen_range(1..(MAX_MESSAGE_LEN * 2));
-            dbg!(random_len);
```
niklaslong, line 1067: Nice 👍

src/tests/resistance/fuzzing.rs
```diff
@@ -398,6 +820,143 @@ async fn fuzzing_slightly_corrupted_version_during_handshake_responder_side() {
     node.stop().await;
 }
 
+#[tokio::test]
+async fn fuzzing_slightly_corrupted_version_when_node_initiates_handshake() {
+    // ZG-RESISTANCE-003 (part 4, version only)
```
niklaslong, line 825: This is clearer 👍

src/tests/resistance/fuzzing.rs
```diff
@@ -1153,3 +2335,43 @@ fn random_non_valid_u32(rng: &mut ChaCha8Rng, value: u32) -> u32 {
...
+            header.checksum = random_non_valid_u32(rng, header.checksum);
+            header.encode(&mut buffer).unwrap();
+
+            buffer
+        })
+        .collect()
+}
+
+/// Picks `n` random messages from `message_pool`, encodes them and corrupts the body-length field
+fn encode_messages_and_corrupt_body_length_field(
```
niklaslong, line 2360: Note: these functions can likely be dedupped (only difference is the field being mutated).

src/tests/resistance/fuzzing.rs
```diff
@@ -1153,3 +2335,43 @@ fn random_non_valid_u32(rng: &mut ChaCha8Rng, value: u32) -> u32 {
...
+    n: usize,
+    message_pool: Vec<Message>,
+) -> Vec<Vec<u8>> {
+    (0..n)
+        .map(|_| {
+            let message = message_pool.choose(rng).unwrap();
+
+            let mut buffer = vec![];
+            let mut header = message.encode(&mut buffer).unwrap();
+            header.body_length = random_non_valid_u32(rng, header.checksum);
```
niklaslong, line 2371:
```suggestion
            header.body_length = random_non_valid_u32(rng, header.body_length);
```

src/tests/resistance/fuzzing.rs
```diff
@@ -398,6 +820,143 @@ async fn fuzzing_slightly_corrupted_version_during_handshake_responder_side() {
...
+                ))
+                .write_to_stream(&mut peer_stream)
+                .await
+                .unwrap();
+                let verack = Message::read_from_stream(&mut peer_stream).await.unwrap();
+                assert_matches!(verack, Message::Verack);
+
+                // send bad version instead of verack
+                let mut rng = peer_rng.lock_owned().await;
+                let version_to_corrupt =
```
niklaslong, line 936: The version constructed above can be reused ☝️

niklaslong (changes requested):
Great work on this! 💪 
I left some comments, most notably regarding the sync primitives being used.

src/tests/resistance/fuzzing.rs
Mirko-von-Leipzig: I added the function because I kept mixing up the argument order.

I agree the name is a bit poor - I suppose conventionally one might implement `TryFrom` instead? Or should I just nuke it completely?

src/tests/resistance/fuzzing.rs
Mirko-von-Leipzig: I'll scope them - thanks

src/tests/resistance/fuzzing.rs
Mirko-von-Leipzig: Good catch - guess I need to re-run those tests :D

src/tests/resistance/fuzzing.rs
Mirko-von-Leipzig: I didn't think the code amount warranted it. I've given it a brief shot, but it all seemed very clumsy. Maybe just a simple Rustsm I'm missing? (or brains dead :D also likely)

src/tests/resistance/fuzzing.rs
Mirko-von-Leipzig: I wasn't aware there were differences - makes sense.

src/tests/resistance/fuzzing.rs
niklaslong: I'd remove it tbh, perhaps just make a comment of the argument order?

src/tests/resistance/fuzzing.rs
niklaslong: Indeed, we'll need to restructure this a little (not in this PR though).

niklaslong (approved):

PR #51, opened 2021-05-20
missing fuzz message tests

Mirko-von-Leipzig:


src/tests/resistance/fuzzing.rs
```diff
@@ -2392,3 +2154,23 @@ fn encode_messages_and_corrupt_body_length_field(
         })
         .collect()
 }
+
+/// Returns the set of messages used for fuzz-testing.
+/// This notably excludes [Message::Version] because it is
+/// usually tested separately.
+fn default_fuzz_messages() -> Vec<Message> {
```
Mirko-von-Leipzig, line 2161: We may want to add variants which aren't just `empty()`, but for now this should suffice.

niklaslong (approved):
Nice 👍

PR #52, opened 2021-05-21
ref: split fuzz tests by payload type

niklaslong:
This moves common code into `resistance/mod.rs` and introduces new files for each fuzzed payload type.


Mirko-von-Leipzig (approved):
Nice!

PR #53, opened 2021-05-21
feat: performance 002 part 1

Mirko-von-Leipzig:


src/tests/performance.rs
```diff
@@ -0,0 +1,234 @@
...
+        .start()
+        .await;
+
+    // start peer event manager
+    let event_manager = tokio::spawn(async move {
+        let mut stats = ConnectionStats::default();
+
+        loop {
+            match event_rx.recv().await.unwrap() {
+                PeerEvent::Rejected => stats.rejected += 1,
```
niklaslong, line 112: Note for the future: if we want prometheus support, we'll likely want to replace these with https://docs.rs/metrics/0.16.0/metrics/.

src/tests/performance.rs
```diff
@@ -0,0 +1,234 @@
...
+struct ConnectionStats {
+    /// count of successfuly handshaken connections
+    pub success: u16,
+    /// count of connections rejected pre-handshake
+    pub rejected: u16,
+    /// count of connections terminated post-handshake
+    pub terminated: u16,
+    /// running count of active connections (approximate since nodes
+    /// are executing concurrently)
+    pub active_count: Vec<u16>,
```
niklaslong, line 30: Why use a `Vec` here?
Have you considered [atomics](https://doc.rust-lang.org/std/sync/atomic/index.html) (this would remove the need for a oneshot channel)?

niklaslong (commented):
Left a few suggestions, otherwise looks good 👍

src/tests/performance.rs
Mirko-von-Leipzig: I used a `Vec` because I wanted to see the change over time. Turned out to be pretty boring for the conventional cases so I ended up just using the max of it. Can be replaced with just the `max` now.

I'll make another pass and simplify it (along with atomics)

Mirko-von-Leipzig: Having another look at this - the oneshot channel will remain, since a peer node needs to remain active until we are done processing *all* the other peers. Each peer therefore needs to remain active until we tell it to shutdown (via the oneshot channel) or its connection gets terminated.

The alternative would be to kill the node (which would in turn terminate the peer connections), but that's not a very clean way of doing it?

Mirko-von-Leipzig: One does need a concurrent process which tracks the number of peer nodes that have completed the test. Once all are done, then we can proceed with the cleanup / shutdown of the peers and node.

Currently this is the `event_manager` task. If we switch to atomics over a channel, then we will require the `event_manager` to loop and:
1. check the atomic counts (to see if all peers are done), exit if true; else
2. sleep a bit
3. repeat

I prefer the message / event based way; but I'm fine with changing it to atomics and a sleep-check loop

niklaslong: In that case, let's keep as is for now 👍

PR #56, opened 2021-05-27
Completed performance test 001 with GetData-Block case

Mirko-von-Leipzig:


src/tests/performance/blocks.rs
```diff
@@ -0,0 +1,222 @@
...
+            Cell::new("Time (s)").set_alignment(CellAlignment::Center),
+            Cell::new("Request/s").set_alignment(CellAlignment::Center),
+        ]);
+
+    // Start node seeded with initial testnet blocks,
+    // with max peers set so that our peers should never be rejected.
+    let mut node: Node = Default::default();
+    node.initial_action(Action::SeedWithTestnetBlocks {
+        socket_addr: new_local_addr(),
+        block_count: 3,
```
niklaslong, line 116: We should test larger block counts in future, 3 blocks isn't very representative of the load an actual sync would put on a node.

src/tests/performance/blocks.rs
```diff
@@ -0,0 +1,222 @@
...
+                })
+                .collect::<VecDeque<_>>();
+
+            peer_handles.push(tokio::spawn(async move {
+                let mut stream = initiate_handshake(node_addr).await.unwrap();
+
+                let filter = MessageFilter::with_all_auto_reply();
+
+                let mut latencies = Vec::with_capacity(REQUESTS);
+                for _ in 0..REQUESTS {
```
niklaslong, line 150: This implies the same requests are made in cycles of 3? I wonder if `zcashd` or `zebra` would want throttle the same request from the same address (as this test runs fine, I'd assume they don't currently).

src/tests/performance/blocks.rs
```diff
@@ -0,0 +1,222 @@
...
+                    let (request, expected) = requests.pop_back().unwrap();
```
niklaslong, line 151: Suggestion: you could use a `Vec` with indices mod 3 of the `0..request` above instead of mutating a queue?

src/tests/performance/blocks.rs
```diff
@@ -0,0 +1,222 @@
...
+    // We test the overall performance of a node's GetData-Block latency.
+    //
+    // Note: This test does not assert any requirements, but requires manual inspection
+    //       of the results table. This is because the results will rely on the machine
+    //       running the test.
+    //
+    // ZCashd: Strange, small slow-down as soon as multiple peers are present.
+    //         Otherwise performs consistently up to around 200 concurrent peers.
+    //
+    // Zebra: Chokes right out of the gate, likely to the DoS behaviour of spamming GetAddr and GetData.
```
niklaslong, line 33: Good find 👍 
This might be indicative of throttling, since we're making the same requests from the same addresses?

niklaslong (commented):
Nice work, left some comments for discussion and some suggestions.

src/tests/performance/blocks.rs
Mirko-von-Leipzig: Yes that's currently the case. Would be trivial to extend that to be random, but with only three blocks the combinations are somewhat limited.

src/tests/performance/blocks.rs
Mirko-von-Leipzig: It's possible, but I couldn't really make sense of what's happening.

It would really help if the logs could be sent to an elastic search type database. The zebra logs are structured (I think), so that would let us filter them. But zcashd looks like vanilla logs, so wouldn't help too much there.

src/tests/performance/blocks.rs
Mirko-von-Leipzig: Agreed.

src/tests/performance/blocks.rs
Mirko-von-Leipzig: Ah man I always forget just using indices lets you have multiple accesses

Mirko-von-Leipzig: Fixed

niklaslong (approved):

PR #58, opened 2021-05-28
fix: conformance test 013

Mirko-von-Leipzig:
Previously tested that inbound addresses were gossiped.

- Now correctly tests that only peer listener addresses get gossiped
- Exercises known Zebra bug: https://github.com/ZcashFoundation/zebra/pull/2120

### DIFF
--- a/src/tests/conformance/messages.rs
+++ b/src/tests/conformance/messages.rs
@@ -564,21 +564,24 @@ async fn eagerly_crawls_network_for_peers() {
     node.stop().await;
 }
 
-// #[tokio::test]
+#[tokio::test]
 async fn correctly_lists_peers() {
     // ZG-CONFORMANCE-013
     //
     // The node responds to a `GetAddr` with a list of peers it’s connected to. This command
     // should only be sent once, and by the node initiating the connection.
     //
-    // Test procedure
-    //  Start a node, and sequentially for each peer `i` of `N`:
-    //      1. Initiate a connection and complete the handshake
-    //      2. Send `GetAddr` request
-    //      3. Receive `Addr` response
-    //      4. Verify `Addr` contains list of previous `i-1` peers
+    // In addition, this test case exercises the known zebra bug: https://github.com/ZcashFoundation/zebra/pull/2120
     //
-    // This test currently fails for zcashd and zebra passes.
+    // Test procedure
+    //      1. Establish N peer listeners
+    //      2. Start node which connects to these N peers
+    //      3. Create i..M new connections which,
+    //          a) Connect to the node
+    //          b) Query GetAddr
+    //          c) Receive Addr == N peer addresses
+    //
+    // This test currently fails for both zcashd and zebra.
     //
     // Current behaviour:
     //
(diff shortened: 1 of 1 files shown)


src/tests/conformance/messages.rs
```diff
@@ -564,67 +564,121 @@ async fn eagerly_crawls_network_for_peers() {
...
     //          the connection. If the node initiates the connection then the command is recoginized,
     //          but likely ignored (because only the initiating node is supposed to send it).
     //
-    //  zebra:  Infinitely spams `GetAddr` and `GetData`. Can be coaxed into responding correctly if
-    //          all its peer connections have responded to `GetAddr` with a non-empty list.
+    //  zebra:  Times out due to infinitely spamming `GetAddr` and `GetData`.
+    //
+    //          Can be coaxed into responding by sending a non-empty Addr in
+    //          response to node's GetAddr. This still fails as it includes previous inbound
+    //          connections in its address book (as in the bug listed above).
```
niklaslong, line 596: Have you tried running with latest `main` or after the bug was supposedly fixed?

src/tests/conformance/messages.rs
```diff
@@ -564,67 +564,121 @@ async fn eagerly_crawls_network_for_peers() {
...
+    let mut inbound = Vec::new();
 
-        Message::GetAddr
-            .write_to_stream(&mut new_peer)
-            .await
-            .unwrap();
+    // Connect to node and request GetAddr.
+    // We perform multiple iterations in order to exercise the above Zebra bug.
+    for i in 0..5 {
+        let mut stream = initiate_handshake(node.addr()).await.unwrap();
```
niklaslong, line 637: the version in `initiate_handshake` might need to send the listener address?

niklaslong (commented):
Great job on this! A few small open questions for consideration.

src/tests/conformance/messages.rs
Mirko-von-Leipzig: I wasn't aware it was in master. I've re-run the test and updated the result.

It now no longer spams, but it remains in "network setup" state until we respond to the `GetAddr` in which case it then responds.

src/tests/conformance/messages.rs
Mirko-von-Leipzig: I'm honestly not sure. In our current test setup I think it doesn't matter - because we are on a local network which bitcoin terms as "unroutable". From the second bitcoin example for [version](https://en.bitcoin.it/wiki/Protocol_documentation#version)

` notice that it does not fill out the address information at all when the source or destination is "unroutable".`

And the send/recv address are simply filled out identically:
```
 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 FF FF 00 00 00 00 00 00 - Recipient address info - see Network Address
 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 FF FF 00 00 00 00 00 00 - Sender address info - see Network Address
```

niklaslong: Fair enough, in normal circumstances, the node might expect to find the peer's listener there. This will get refactored when adding `pea2pea` so no need to fix this now.

niklaslong (approved):

PR #60, opened 2021-05-31
feat: synthetic nodes

niklaslong:
This PR:
- adds `pea2pea` backed synthetic nodes abstracting handshake (optional), read, write and message filtering.
- refactors and restructures `message.rs` into modules.
- removes custom `assert_matches!` in favour of [a well supported dep](https://github.com/murarth/assert_matches) as it supports guards.


src/helpers/synthetic_peers.rs
```diff
@@ -0,0 +1,180 @@
...
+    net::SocketAddr,
+};
+
+pub struct SyntheticNode {
+    inner_node: InnerNode,
+    inbound_rx: Receiver<Message>,
+}
+
+impl SyntheticNode {
+    pub fn new(node: Node, enable_handshaking: bool, message_filter: MessageFilter) -> Self {
```
Mirko-von-Leipzig, line 25: Is there a way one can re-configure this on the fly?

I'm not sure we have a use-case currently, but I'm thinking of a situtation where one wants to change `message_filter` in-between receive calls. I suppose that would also be impossible to control correctly since `Pea2Pea` starts a background task for reading the stream (?).

src/helpers/synthetic_peers.rs
```diff
@@ -0,0 +1,180 @@
...
+            inner_node.enable_handshaking();
+        }
+
+        Self {
+            inner_node,
+            inbound_rx: rx,
+        }
+    }
+
+    pub async fn connect(&self, target: SocketAddr) -> Result<()> {
```
Mirko-von-Leipzig, line 44: An alternate API to consider:

1. Have a SynthNodeBuilder (which has no recv/send methods)
2. Has a variety of `build` type methods, which return a `SynthNode`
     1. `connect` (connect, without any handshaking)
     2. `accept` (listen and accept without handshaking)
     3. `initiate_handshake` (connect and handshake)
     4. `recieve_handshake` (listen, accept and handshake)
     5. ... 

This prevents forgetting to connect (since one cannot receive / send on the builder). But could be too much boiler plate for not much more upside (:

src/helpers/synthetic_peers.rs
niklaslong: We could mutate the `SyntheticNode` if this is necessary (tough we'd need to `Arc` the inner node).

src/helpers/synthetic_peers.rs
niklaslong: Or did you mean as a replacement for `SyntheticNode::new`?

Mirko-von-Leipzig: Yes, as a replacement for `new`. All cases are currently covered, but one can misuse by doing

```
let peer = SyntheticNode ::new(...);
peer.recv_message(); // will panic!("all receivers dropped");
```

I'm suggesting something along these lines

```
let peer = SyntheticNode::builder(...)
                    .initiate_handshake(node_addr).await.unwrap();
peer.recv_message();
```

but its mostly a trivial thing, just means you can't forget to call `connect` etc.

niklaslong: I'd actually expect that first example to simply hang, since it will just `await` on the channel `recv`.

Mirko-von-Leipzig (approved):
Looking good!

PR #62, opened 2021-06-03
resistance 006

Mirko-von-Leipzig:
This is not quite what you had in mind, but since it took some effort I thought I'd rather you look at it.

A lot of the code is just stats related, the core of it is simply

Spin up N concurrent peers, which
1. send 100 requests, and validate the reply
2. send a corrupt message, and check the node's reaction

Since this is happening concurrently, the node should be reacting to valid and corrupt messages alike.


src/tests/resistance/fuzzing_flood.rs
```diff
@@ -0,0 +1,640 @@
...
+    // ZG-RESISTANCE-006
+    //
+    // Simulate high load situations, while also sending corrupt messages.
+    //
+    // We simulate active peers which send 100 requests (and validate the replies), and finally send a corrupt message
+    // which ends the connection. We measure the latency of the replies and the type of response to the corrupt message.
+    //
+    //  *NOTE* run with `cargo test --release tests::resistance::fuzzing_flood::rising_fuzz -- --nocapture`
+    //
+    // Currently only works for zcashd as it requires block seeding.
```
niklaslong, line 85: Have you tested the new zebra seeding with this yet?

src/tests/resistance/fuzzing_flood.rs
```diff
@@ -0,0 +1,640 @@
...
+    //
+    // Example zcashd run result (notably some connections appear to hang after sending the corrupt message, and some
+    // get replies to the corrupt message).
+    //
+    // Stats
```
niklaslong, line 90: Nice! 👌

src/tests/resistance/fuzzing_flood.rs
```diff
@@ -0,0 +1,640 @@
...
+    // ╞═══════╪══════════╪════════════╪═══════════╪═══════════╪═══════════╪════════════╪════════════╪══════════╪═════════╪═════════╪═══════════╪═════════════╪════════════╪══════════╡
+    // │     1 ┆      100 ┆          1 ┆         1 ┆         0 ┆         0 ┆          0 ┆          1 ┆        0 ┆       0 ┆       0 ┆         0 ┆           0 ┆          0 ┆     0.03 │
+    // ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┤
+    // │    10 ┆      100 ┆         10 ┆        10 ┆         0 ┆         0 ┆          1 ┆          5 ┆        0 ┆       1 ┆       1 ┆         0 ┆           0 ┆          2 ┆    10.72 │
+    // ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┤
+    // │    20 ┆      100 ┆         18 ┆        20 ┆         0 ┆         0 ┆          0 ┆         14 ┆        0 ┆       3 ┆       1 ┆         0 ┆           0 ┆          2 ┆    21.94 │
```
niklaslong, line 99: What's "max active", I notice it's lower than the peer count/accepted handshakes in some cases? Are these expected count - hung at the peak?

src/tests/resistance/fuzzing_flood.rs
```diff
@@ -0,0 +1,640 @@
...
+    // │   800 ┆      100 ┆        766 ┆       800 ┆         0 ┆         0 ┆          4 ┆        543 ┆       25 ┆     101 ┆      34 ┆         0 ┆           0 ┆         93 ┆   187.49 │
+    // ╰───────┴──────────┴────────────┴───────────┴───────────┴───────────┴────────────┴────────────┴──────────┴─────────┴─────────┴───────────┴─────────────┴────────────┴──────────╯
+
+    // Request latencies
+    // ╭───────┬──────────┬──────────┬──────────┬─────────────┬──────────┬──────────┬──────────┬──────────┬──────────┬───────────╮
+    // │ Peers ┆ Requests ┆ Min (ms) ┆ Max (ms) ┆ stddev (ms) ┆ 10% (ms) ┆ 50% (ms) ┆ 75% (ms) ┆ 90% (ms) ┆ 99% (ms) ┆ Request/s │
+    // ╞═══════╪══════════╪══════════╪══════════╪═════════════╪══════════╪══════════╪══════════╪══════════╪══════════╪═══════════╡
+    // │     1 ┆      100 ┆        0 ┆        0 ┆           0 ┆        0 ┆        0 ┆        0 ┆        0 ┆        0 ┆   3500.09 │
+    // ├╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌┼╌╌╌╌╌╌╌╌╌╌╌┤
+    // │    10 ┆      100 ┆        0 ┆       52 ┆          19 ┆        0 ┆        0 ┆        1 ┆       50 ┆       51 ┆     37.78 │
```
niklaslong, line 134: That's a big drop!

src/tests/resistance/fuzzing_flood.rs
```diff
@@ -0,0 +1,640 @@
...
+
+    let mut rng = seeded_rng();
+    let valid_pool = valid_queries_responses();
+
+    let mut node: Node = Default::default();
+    node.initial_action(Action::SeedWithTestnetBlocks {
+        socket_addr: new_local_addr(),
+        block_count: 3,
+    })
+    .max_peers(peer_counts.iter().max().unwrap() * 2 + 10)
```
niklaslong, line 219: Is this arbitrarily higher than the peer count?

niklaslong (commented):
Nice work! I wonder if there's anyway of extracting the stat logic (we use it in quite a few tests). I'd probably wait until after the synth node refactor is in as it changes the game from that perspective.

src/tests/resistance/fuzzing_flood.rs
Mirko-von-Leipzig: I suspect its just random chance - its possible that the query chosen for the first test case with `N=1` was something trivial like `Ping` - which should go quicker.

Certainly worth investigating though.

src/tests/resistance/fuzzing_flood.rs
Mirko-von-Leipzig: Yes - just to ensure that the node should technically accept all of peer connections.

src/tests/resistance/fuzzing_flood.rs
Mirko-von-Leipzig: No, I didn't want to cross the two PR's - I can certainly try that though; would be interesting.

src/tests/resistance/fuzzing_flood.rs
Mirko-von-Leipzig: 'max active' is the peak number of concurrent connections. Maybe `peak` is a better word then :D The table headers in general could use an improvement.

Mirko-von-Leipzig:
> 
> 
> Nice work! I wonder if there's anyway of extracting the stat logic (we use it in quite a few tests). I'd probably wait until after the synth node refactor is in as it changes the game from that perspective.

Its worth considering - as you've mentioned before; might make sense to use a proper stats / events crate for eventual ingestion into grafana etc.

src/tests/resistance/fuzzing_flood.rs
Mirko-von-Leipzig: The hung connections and replies to the corrupt message may need some investigating.

Mirko-von-Leipzig: Completed my investigations.

1. Replies were as a result of letting valid messages into the test set
2. Hung connections are the node waiting for more data due to getting a corrupt message length. This is likely the correct behaviour.

Mirko-von-Leipzig:
I've updated the PR with quite a few changes, almost exclusively outside the test case itself.

Notably
1. Replace `panic` with `Err`
2. Cap `VarInt` and `VarStr` lengths
3. No longer `panic` when decoding invalid timestamps

src/protocol/message/mod.rs
```diff
@@ -168,7 +168,12 @@ impl Message {
             MEMPOOL_COMMAND => Self::MemPool,
             TX_COMMAND => Self::Tx(Tx::decode(bytes)?),
             REJECT_COMMAND => Self::Reject(Reject::decode(bytes)?),
-            _ => unimplemented!(),
+            cmd => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::InvalidInput,
```
niklaslong, line 173: 👌

niklaslong (approved):
Great work investigating this! 🎉 
I'm thinking this test (and the suite in general) might benefit from having stats implemented with the `metrics` crate.

PR #63, opened 2021-06-04
zebra seeding

Mirko-von-Leipzig:
I've managed to get zebra seeding going. It was quite annoying and I'm not sure we want to actually do this. I've opened this as a draft to get some thoughts and have a discussion. Also note that I haven't cleaned up any of this yet - so it still has a bit of work left, but I'm going to leave that until we decide.

It has to work around the zebra network setup bug - it spam's `GetAddr` until it has at least 4 (?) active peers.

I've put some [notes down here](https://hackmd.io/@HMoSWBonQLy2g2Lwrz9SoQ/HyDfDaZIu#Zebra-difficult-to-seed-with-blocks-due-bad-fan-out-logic-)

I used the `get_blocks` test case to exercise it. And unfortunately, even though the node is now seeded with data as soon as there aren't 4+ peers, it resumes the `GetAddr` spam and therefore doesn't respond to data queries. So it still fails.

### DIFF
--- a/src/protocol/message/mod.rs
+++ b/src/protocol/message/mod.rs
@@ -173,6 +173,30 @@ impl Message {
 
         Ok(message)
     }
+
+    pub fn command(&self) -> [u8; 12] {
+        match self {
+            Message::Version(_) => VERSION_COMMAND,
+            Message::Verack => VERACK_COMMAND,
+            Message::Ping(_) => PING_COMMAND,
+            Message::Pong(_) => PONG_COMMAND,
+            Message::GetAddr => GETADDR_COMMAND,
+            Message::Addr(_) => ADDR_COMMAND,
+            Message::GetHeaders(_) => GETHEADERS_COMMAND,
+            Message::Headers(_) => HEADERS_COMMAND,
+            Message::GetBlocks(_) => GETBLOCKS_COMMAND,
+            Message::Block(_) => BLOCK_COMMAND,
+            Message::GetData(_) => GETDATA_COMMAND,
+            Message::Inv(_) => INV_COMMAND,
+            Message::NotFound(_) => NOTFOUND_COMMAND,
+            Message::MemPool => MEMPOOL_COMMAND,
+            Message::Tx(_) => TX_COMMAND,
+            Message::Reject(_) => REJECT_COMMAND,
+            Message::FilterLoad(_) => todo!(),
+            Message::FilterAdd(_) => todo!(),
+            Message::FilterClear => todo!(),
+        }
+    }
 }
 
 fn checksum(bytes: &[u8]) -> u32 {
--- a/src/setup/node.rs
+++ b/src/setup/node.rs
@@ -1,10 +1,14 @@
 use crate::{
     helpers::respond_to_handshake,
     protocol::{
-        message::filter::{Filter, MessageFilter},
+        message::{
+            filter::{Filter, MessageFilter},
+            Message,
+        },
         payload::{
-            block::{Block, Headers},
-            Hash, Inv, Nonce,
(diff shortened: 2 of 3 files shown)


src/setup/node.rs
```diff
@@ -165,16 +170,110 @@ impl Node {
...
+                }
+
+                let addr = self.addr();
+
+                let mut handles = Vec::new();
+
+                // Check which blocks we have sent, so that we can know when we are done.
+                let blocks_sent = Arc::new(Mutex::new(vec![false; block_count]));
+
+                // Spin up four concurrent peers - this seems to be the magic number to get Zebra out of
```
niklaslong, line 189: Weird but cool you discovered this!

src/tests/conformance/messages.rs
```diff
@@ -694,13 +694,16 @@ async fn get_blocks() {
...
     .start()
     .await;
 
+    println!("Starting test!");
+
     let blocks = Block::initial_testnet_blocks();
 
     let mut stream = initiate_handshake(node.addr()).await.unwrap();
-    let filter = MessageFilter::with_all_auto_reply();
+    let filter = MessageFilter::with_all_auto_reply().enable_logging();
```
niklaslong, line 706:
```suggestion
    let filter = MessageFilter::with_all_auto_reply();
```

src/tests/conformance/messages.rs
```diff
@@ -694,13 +694,16 @@ async fn get_blocks() {
         socket_addr: new_local_addr(),
         block_count: 3,
     })
+    // .log_to_stdout(true)
```
niklaslong, line 697:
```suggestion

```

src/tests/conformance/messages.rs
```diff
@@ -694,13 +694,16 @@ async fn get_blocks() {
...
     .start()
     .await;
 
+    println!("Starting test!");
```
niklaslong, line 701:
```suggestion
```

niklaslong (commented):
Looking good! Made a few suggestions :)

src/setup/node.rs
```diff
@@ -165,16 +170,110 @@ impl Node {
...
+                if block_count == 0 {
+                    return;
+                }
+
+                let addr = self.addr();
+
+                let mut handles = Vec::new();
+
+                // Check which blocks we have sent, so that we can know when we are done.
+                let blocks_sent = Arc::new(Mutex::new(vec![false; block_count]));
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: The actual block does matter - the node can send the same query to multiple of the peers. So the same block gets sent to the node multiple times. Its quite timing dependent.

src/setup/node.rs
```diff
@@ -13,9 +17,10 @@ use crate::{
 use tokio::{
     net::TcpListener,
     process::{Child, Command},
+    sync::Mutex,
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: At the time I thought I ran into this issue: https://github.com/rust-lang/rust/issues/71072

But I think it was just user error, and I forgot to change it back. Will fix.

src/protocol/message/mod.rs
```diff
@@ -173,6 +173,30 @@ impl Message {
 
         Ok(message)
     }
+
+    pub fn command(&self) -> [u8; 12] {
+        match self {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I think this is the inverse mapping of those

Mirko-von-Leipzig:
Closing, will keep branch for review with Zebra purposes.

PR #64, opened 2021-06-06
ref: handshake tests with synthetic nodes

niklaslong:
Note: both `zcashd` and `zebra` only close their write half of the connection and don't fully drop the socket when a disconnect is expected (pea2pea doesn't detect the stream close at time of writing: https://github.com/ljedrz/pea2pea/issues/5). The previous version of the tests detected a disconnect because we were using `read_exact` instead of `read` as does pea2pea. I've also updated the spec and pass/fail table to reflect the changes.

Interestingly both `zcashd` and `zebra` also have failing ping/pong tests (if we actually check the `nonce`). 

The synthetic node API is also unstable at this point, though inching closer to something I'm happy with.

### DIFF
--- a/src/helpers/synthetic_peers.rs
+++ b/src/helpers/synthetic_peers.rs
@@ -7,14 +7,21 @@ use assert_matches::assert_matches;
 use pea2pea::{
     connections::ConnectionSide,
     protocols::{Handshaking, Reading, Writing},
-    Connection, Node, Pea2Pea,
+    Connection, Node, NodeConfig, Pea2Pea,
 };
 use tokio::sync::mpsc::{self, Receiver, Sender};
 
 use std::{
     io::{Cursor, ErrorKind, Result},
-    net::SocketAddr,
+    net::{IpAddr, Ipv4Addr, SocketAddr},
 };
+
+#[derive(Default)]
+pub struct SyntheticNodeConfig {
+    pub network_config: Option<NodeConfig>,
+    pub enable_handshaking: bool,
+    pub message_filter: MessageFilter,
+}
 
 /// Conventient abstraction over the `pea2pea`-backed node to be used in tests.
 pub struct SyntheticNode {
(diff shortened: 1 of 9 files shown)


src/helpers/synthetic_peers.rs
```diff
@@ -131,7 +212,7 @@ impl Reading for InnerNode {
         let header = MessageHeader::decode(&mut Cursor::new(header_bytes))?;
 
         // Check buffer contains the announce message lenght.
```
Mirko-von-Leipzig, line 214:
```suggestion
        // Check buffer contains the announced message length.
```

src/tests/conformance/handshake.rs
```diff
@@ -223,96 +191,84 @@ async fn reject_non_version_replies_to_version() {
...
     let genesis_block = Block::testnet_genesis();
     let block_hash = genesis_block.double_sha256().unwrap();
     let block_inv = Inv::new(vec![genesis_block.inv_hash()]);
     let block_loc = LocatorHashes::new(vec![block_hash], Hash::zeroed());
-    let mut test_messages = vec![
-        Message::Version(Version::new(new_local_addr(), new_local_addr())),
+    let test_messages = vec![
         Message::GetAddr,
         Message::MemPool,
+        Message::Verack,
```
Mirko-von-Leipzig, line 271: This probably shouldn't be here for this test case?

```suggestion
```

Mirko-von-Leipzig (changes requested):
Couple of minor changes.

src/tests/conformance/handshake.rs
niklaslong: Good catch!

src/helpers/synthetic_peers.rs
```diff
@@ -211,7 +211,7 @@ impl Reading for InnerNode {
         let header_bytes = &buffer[..HEADER_LEN];
         let header = MessageHeader::decode(&mut Cursor::new(header_bytes))?;
 
-        // Check buffer contains the announce message lenght.
+        // Check buffer contains the announce message length.
```
Mirko-von-Leipzig, line 214:
```suggestion
        // Check buffer contains the announced message length.
```

Mirko-von-Leipzig (changes requested):
Still a small typo left I think?

Mirko-von-Leipzig (approved):

PR #65, opened 2021-06-07
deps: update pea2pea

niklaslong:
This PR also slightly modifies the filtering used by the synthetic nodes.


Mirko-von-Leipzig (approved):
LGTM

PR #66, opened 2021-06-08
fix: fuzz generators

Mirko-von-Leipzig:
They were incorrectly encoding messages as `<body><header>`.

### DIFF
--- a/src/tests/resistance/fuzzing_incorrect_checksum.rs
+++ b/src/tests/resistance/fuzzing_incorrect_checksum.rs
@@ -1,7 +1,7 @@
 use crate::{
     helpers::{autorespond_and_expect_disconnect, initiate_handshake, initiate_version_exchange},
     protocol::{
-        message::Message,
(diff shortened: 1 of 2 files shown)


src/tests/resistance/fuzzing_incorrect_checksum.rs
```diff
@@ -368,8 +368,10 @@ fn encode_messages_and_corrupt_checksum(
         .map(|_| {
             let message = message_pool.choose(rng).unwrap();
 
-            let mut buffer = vec![];
-            let mut header = message.encode(&mut buffer).unwrap();
+            let mut body_buffer = Vec::new();
+            let mut header = message.encode(&mut body_buffer).unwrap();
+
+            let mut buffer = Vec::with_capacity(body_buffer.len() + 24);
```
niklaslong, line 374: fyi, there's a constant for various message properites including header length here: https://github.com/eqlabs/ziggurat/blob/master/src/protocol/message/constants.rs

PR #68, opened 2021-06-09
feat/metrics

Mirko-von-Leipzig:
This is PoC for switching to `metrics` crate for our performance related stats.

Using a standard crate will allow easy integration with `promotheus` and `graphana` like services.

PoC includes a custom `SimpleRecorder` which aggregates the stats, and allows access to them. This is implemented as a singleton ala `lazy_static`. While not ideal, this is the model used by `metrics` which only allows for a single active recorder at a time.

The `ping` performance test was used to exercise the stat aggregation.

Also of interest - zebra now performs really well in this test (see the updated results table).

Looking for feedback, before proceeding to do docs, cleanup and refactoring of other performance stats.


niklaslong (approved):
Nice! This is great! 🎉 
Just a thought, is there a way to make the tables generic (especially if we're reusing the same ones between tests)?

Mirko-von-Leipzig:
> 
> 
> Nice! This is great! 🎉
> Just a thought, is there a way to make the tables generic (especially if we're reusing the same ones between tests)?

I will have to try it and see. I think it might be easier to switch to: https://lib.rs/crates/tabled. 

It creates a table based on `Vec<struct>`, with a row for each struct. This would do the `generic` part for us.

Mirko-von-Leipzig:
Updated PR to use `tabled` to generate tables from structs.

Mirko-von-Leipzig:
There are one or two stat columns that got purged along the way. Mainly because tracking them using metrics is cumbersome. I don't think they're a big loss.

src/tests/performance/connections.rs
```diff
@@ -55,168 +60,202 @@ async fn incoming_active_connections() {
...
-                    // there will always be a `last` item here.
-                    let prev_count = *stats.active_count.last().unwrap();
-                    stats.active_count.push(prev_count - 1);
+    for peers in peer_counts {
+        let test_start = tokio::time::Instant::now();
+        // channel for peer event management (ensure buffer is more than large enough)
+        let (event_tx, mut event_rx) = tokio::sync::mpsc::channel::<PeerEvent>(peers as usize * 3);
+
+        // start peer event manager
+        let event_manager = tokio::spawn(async move {
```
niklaslong, line 121: We may not need this as the stats are protected by a mutex? This approach (without the mutex) may be slightly faster but I think simplicity is probably the best thing to strive for in tests?

src/tests/performance/connections.rs
```diff
@@ -14,34 +17,36 @@ enum PeerEvent {
...
-#[derive(Default, Debug)]
-struct ConnectionStats {
-    /// count of successfuly handshaken connections
-    pub success: u16,
-    /// count of connections rejected pre-handshake
+#[derive(Tabled, Default, Debug, Clone)]
+struct Stats {
+    #[header(" N ")]
+    pub max_peers: u16,
+    #[header(" M ")]
```
niklaslong, line 24: Let's use the full name as a header (would also be more consistent with our other tables)?

niklaslong (approved):
Made a few suggestions but otherwise looks good—great work on this! 🎉

src/tests/performance/connections.rs
Mirko-von-Leipzig: I either forgot to `metricify` or overwrote my changes here :/

Mirko-von-Leipzig: I must have missed this test case when updating for `metrics`

src/tests/performance/connections.rs
Mirko-von-Leipzig: Done

niklaslong (approved):

PR #73, opened 2021-06-23
refactor:  fuzz tests now use SynthNode

Mirko-von-Leipzig:
Updates the fuzz tests to use the `SyntheticNode` and its builder pattern.

In addition, the conformance tests also now use the builder pattern.

This also let us deprecate the original node constructor and various helper functions.


src/tests/resistance/fuzzing_corrupted_messages.rs
```diff
@@ -102,43 +117,51 @@ async fn corrupted_version_when_node_initiates_handshake() {
 
     let locked_rng = Arc::new(RwLock::new(seeded_rng()));
 
-    // create tcp listeners for peer set (port is only assigned on tcp bind)
-    let mut listeners = Vec::with_capacity(ITERATIONS);
+    // create peers (we need their ports to give to the node)
+    let mut peers = Vec::with_capacity(ITERATIONS);
     for _ in 0..ITERATIONS {
-        listeners.push(TcpListener::bind(new_local_addr()).await.unwrap());
+        let peer = SyntheticNode::builder()
```
niklaslong, line 123: `build_n` here?

src/tests/resistance/fuzzing_corrupted_messages.rs
```diff
@@ -169,47 +192,53 @@ async fn corrupted_version_inplace_of_verack_when_node_initiates_handshake() {
...
```
niklaslong, line 198: `build_n`? (there are a bunch more too)

src/tests/resistance/fuzzing_corrupted_messages.rs
```diff
@@ -169,47 +192,53 @@ async fn corrupted_version_inplace_of_verack_when_node_initiates_handshake() {
...
-                ));
-                version.write_to_stream(&mut peer_stream).await.unwrap();
-                let verack = Message::read_from_stream(&mut peer_stream).await.unwrap();
+                // Receive verack
+                let (_, verack) = peer.recv_message().await;
                 assert_matches!(verack, Message::Verack);
 
-                // send bad version instead of verack
+                // Send bad version instead of verack
+                let version = Message::Version(Version::new(node_addr, peer.listening_addr()));
```
niklaslong, line 229: I think this naming is slightly confusing, perhaps `node` and `synthetic_node`/`synth_node`/ `s_node` would be better? (I think we should actually choose a convention we stick with throughout the codebase).

src/tests/resistance/fuzzing_incorrect_checksum.rs
```diff
@@ -97,20 +113,31 @@ async fn version_with_incorrect_checksum_during_handshake_responder_side() {
...
+        let version = Message::Version(Version::new(node.addr(), peer.listening_addr()));
+        let mut body_buffer = Vec::new();
+        let mut header = version.encode(&mut body_buffer).unwrap();
 
         // Set the checksum to a random value which isn't the current value.
         header.checksum = random_non_valid_u32(&mut rng, header.checksum);
 
-        let _ = header.write_to_stream(&mut peer_stream).await;
-        let _ = peer_stream.write_all(&message_buffer).await;
+        let mut buffer = Vec::new();
```
niklaslong, line 131:
```suggestion
        let mut buffer = Vec::with_capacity(header length + body length);
```

src/tests/resistance/fuzzing_incorrect_checksum.rs
```diff
@@ -128,20 +155,31 @@ async fn version_with_incorrect_checksum_post_handshake() {
...
```
niklaslong, line 173:
```suggestion
        let mut buffer = Vec::with_capacity(...);
```

src/tests/resistance/fuzzing_incorrect_checksum.rs
```diff
@@ -328,19 +386,31 @@ async fn incorrect_checksum_post_handshake() {
...
-        let mut message_buffer = vec![];
-        let mut header = message.encode(&mut message_buffer).unwrap();
+        let mut body_buffer = Vec::new();
+        let mut header = message.encode(&mut body_buffer).unwrap();
 
         // Set the checksum to a random value which isn't the current value.
         header.checksum = random_non_valid_u32(&mut rng, header.checksum);
 
-        let mut peer_stream = initiate_handshake(node.addr()).await.unwrap();
+        let mut buffer = Vec::new();
```
niklaslong, line 395:
```suggestion
        let mut buffer = Vec::with_capacity(...);
```

niklaslong (changes requested):
Great work on this, I made a few suggestions 🎉

niklaslong:
I ran this locally and I'm seeing a bunch of errors that aren't timeouts (buffer length,`SetRecorderError`).

src/tests/resistance/fuzzing_corrupted_messages.rs
Mirko-von-Leipzig: I initially just refactored whatever there was to the builder pattern. What we can also do is move the builder outside the loop and then just call `build` inside. Alternatively we do `build_n` outside. I think I prefer just `build` for these sequential tests.

Will refactor these.

Mirko-von-Leipzig:
> 
> 
> I ran this locally and I'm seeing a bunch of errors that aren't timeouts (buffer length,`SetRecorderError`).

The buffer lengths mean I need to set the capacity somewhere - probably the `zeroes` and `random_bytes` tests.

`SetRecorderError` surprises me since that should only occur if the recorder was already set before. Are you running the tests in one big group? Or one at a time? (I've only ever run them one at a time, so it may be that I need to do something different here).

niklaslong:
> Are you running the tests in one big group? Or one at a time?

In a group but sequentially: `cargo t -- --test-threads 1`

src/tests/resistance/fuzzing_corrupted_messages.rs
Mirko-von-Leipzig: Yeah good idea. `synth_node` and `node` sound good to me?

niklaslong (approved):

PR #74, opened 2021-06-23
fix: connection performance test

Mirko-von-Leipzig:
The test now enforces a timeout (which fixes Zebra hanging).

Now it causes connection bind failures instead, which get logged as
errors in the stats.


niklaslong (approved):
LGTM 👍

PR #75, opened 2021-06-24
API docs and small refactors

niklaslong:
This PR is part of the ongoing effort to polish the APIs and docs.

Notable changes:
- rename `helpers` -> `tools`
- rename `synthetic_peers` -> `synthetic_node`
- refactor and move  message filtering into `tools`


src/tests/conformance/messages.rs
```diff
@@ -77,7 +75,7 @@ async fn reject_invalid_messages() {
     // List of test messages and their expected Reject kind.
     let cases = vec![
         (
-            Message::Version(Version::new(node.addr(), new_local_addr())),
+            Message::Version(Version::new(node.addr(), "0.0.0.0:0".parse().unwrap())),
```
niklaslong, line 78: This is equivalent to the previous code, just removed the helper. We may want to include a different address here in future but it will require refactoring the test.

src/vectors/mod.rs
```diff
@@ -1,34 +1,33 @@
...
+    /// Testnet block at height `280000` (first sapling).
     pub static ref BLOCK_TESTNET_280000_BYTES: Vec<u8> =
         <Vec<u8>>::from_hex(include_str!("block-test-0-280-000.txt").trim()).unwrap();
+    /// Testnet block at height `584000` (first blossom).
     pub static ref BLOCK_TESTNET_584000_BYTES: Vec<u8> =
         <Vec<u8>>::from_hex(include_str!("block-test-0-584-000.txt").trim()).unwrap();
+    /// Testnet block at height `903800` (first heartwood).
     pub static ref BLOCK_TESTNET_903800_BYTES: Vec<u8> =
         <Vec<u8>>::from_hex(include_str!("block-test-0-903-800.txt").trim()).unwrap();
+    /// Testnet block at height `1028500` (first canopy).
```
Mirko-von-Leipzig, line 30: I wonder if we don't want to also have an alias for these? Like `BLOCK_TESTNET_CANOPY_0`? Maybe not here, but in the `Block::canopy_0()`? Eh.. not required.

src/tools/synthetic_node.rs
```diff
@@ -25,12 +26,31 @@ use std::{
...
+    fmt()
+        .with_test_writer()
+        .with_env_filter(EnvFilter::from_default_env())
+        .init();
+}
+
+/// Describes the handshake to be performed by a [`SyntheticNode`].
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum Handshake {
+    /// [`Version`] and [`Verack`] in both directions.
```
Mirko-von-Leipzig, line 42: Thanks for documenting this (and all the rest).

src/tools/synthetic_node.rs
```diff
@@ -119,13 +138,14 @@ impl SyntheticNodeBuilder {
...
 
-/// Conventient abstraction over the `pea2pea`-backed node to be used in tests.
+/// Conventient abstraction over a `pea2pea` node.
 pub struct SyntheticNode {
     inner_node: InnerNode,
     inbound_rx: Receiver<(SocketAddr, Message)>,
 }
 
 impl SyntheticNode {
+    // FIXME: remove in favour of calling `SyntheticNodeBuilder::default()` or `new` directly?
```
Mirko-von-Leipzig, line 148: I like this way around since it makes explicit that this is a builder for `SyntheticNode`. Maybe even removing the `Default` for the builder and make it a private `new()` - so the only way of constructing one is via `SyntheticNode::builder()`?

It's not a strong opinion though, so happy for it go either way.

src/tools/message_filter.rs
```diff
@@ -0,0 +1,135 @@
+//! Message filtering types and utilities.
```
Mirko-von-Leipzig, line 1: I'm just assumming this was copied verbatim? i.e. moved

Mirko-von-Leipzig (approved):
LGTM, some comments but no suggestions. Great work, especially on the nitty documentation.

src/tools/message_filter.rs
niklaslong: I removed some functions and the logging which became unused as a result of the synth node refactor, functionality is much the same.

PR #76, opened 2021-06-28
refactor: performance tests

Mirko-von-Leipzig:
Simplification and renaming of performance tests, and updated results to our target versions.


src/tests/performance/mod.rs
```diff
@@ -88,3 +86,16 @@ pub fn table_float_display(x: &f64) -> String {
 pub fn duration_as_ms(duration: Duration) -> f64 {
     duration.as_millis() as f64
 }
+
+/// Formats a [Table] with our style:
+///  - [pseudo style](Style) (todo - fix this link)
```
Mirko-von-Leipzig, line 91: I want this to link to [Style::pseudo()](https://docs.rs/tabled/0.2.1/tabled/style/struct.Style.html#method.pseudo), but I don't know how to link to an external crate's function?

niklaslong: Should be just [pseudo style](https://docs.rs/tabled/0.2.1/tabled/style/struct.Style.html#method.pseudo) no?

src/tests/performance/ping_pong.rs
```diff
@@ -114,95 +123,85 @@ async fn throughput() {
...
         // add stats to table display
         table.add_row(RequestStats::new(
-            peers as u16,
+            synth_count as u16,
             PINGS,
             latencies,
             time_taken_secs,
         ));
+
+        println!("{}", table);
```
niklaslong, line 166: Suggestion: only print the table at the end of the run?

Mirko-von-Leipzig: Debug aid left behind :D will remove.

src/tests/performance/mod.rs
Mirko-von-Leipzig: That works; but I was hoping for something a bit more automated? (i.e. link to the version we are actually running). What bothers me is that `(Style)` works just fine, but I can't get any further.

niklaslong: Oh you mean the method? What about `[...](Struct::pseudo)`?

niklaslong (approved):

PR #78, opened 2021-06-29
Improve setup code error handling

niklaslong:
Removes `unwrap` in various `Node` functions in favour of returning an `io::Result`. 
Also renames `message::io` -> `message::stream_io` for clarity.

### DIFF
--- a/src/tools/synthetic_node.rs
+++ b/src/tools/synthetic_node.rs
@@ -21,7 +21,7 @@ use tokio::{
 use tracing::*;
 
 use std::{
-    io::{Cursor, Error, ErrorKind, self},
+    io::{self, Cursor, Error, ErrorKind},
     net::{IpAddr, Ipv4Addr, SocketAddr},
     time::Duration,
 };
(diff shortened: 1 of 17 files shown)


src/tools/synthetic_node.rs
```diff
@@ -437,18 +441,18 @@ impl Writing for InnerNode {
...
 }
 
 #[async_trait::async_trait]
 impl Handshaking for InnerNode {
-    async fn perform_handshake(&self, mut conn: Connection) -> Result<Connection> {
+    async fn perform_handshake(&self, mut conn: Connection) -> io::Result<Connection> {
         match (self.handshake, !conn.side) {
             (Some(Handshake::Full), ConnectionSide::Initiator) => {
-                // Possible bug: running zebra node results in internal pea2pea panics:
+                // Possible bug: running zebra node io::Results in internal pea2pea panics:
```
Mirko-von-Leipzig, line 455: Heh - I think search and replace typo here?

Mirko-von-Leipzig (changes requested):
LGTM, just the one search+replace typo.

src/tools/synthetic_node.rs
niklaslong: indeed, good catch!

Mirko-von-Leipzig (approved):

PR #79, opened 2021-06-29
ref: DRY version fuzzing

Mirko-von-Leipzig:
Added Version to default fuzz messages, which allows us to de-duplicated the version specific tests.


niklaslong (approved):
Nice and red 🔥

PR #80, opened 2021-06-29
ref: config error handling

niklaslong:
Removes `unwrap` in various config utility functions. Also addresses a few clippy lints on test targets as a fly-by.


Mirko-von-Leipzig (approved):
LGTM, nice!

PR #81, opened 2021-06-29
ci: include tests in clippy checks

niklaslong:
The fixes for the failing lints are included in #80.


Mirko-von-Leipzig (approved):
Nice! It's bitten me a few times.

PR #85, opened 2021-07-02
feat: automated node config path args

Mirko-von-Leipzig:
The user previously had to specify the location of our generated config file. This change removes this requirement -- since we generate it, we should also pass on the information to the node.

### DIFF
--- a/src/setup/config.rs
+++ b/src/setup/config.rs
@@ -11,6 +11,10 @@ use std::{
 
 use crate::setup::node::Action;
 
+// The names of the files the node configurations will be written to.
+const ZEBRA_CONFIG: &str = "zebra.toml";
+const ZCASHD_CONFIG: &str = "zcash.conf";
+
 const CONFIG: &str = "config.toml";
 const DEFAULT_PORT: u16 = 8080;
 
@@ -65,6 +69,16 @@ pub(super) enum NodeKind {
     Zcashd,
 }
 
+impl NodeKind {
+    /// Path to the configuration file for this [NodeKind]
+    pub(super) fn config_filepath(&self) -> std::path::PathBuf {
+        match self {
+            NodeKind::Zebra => std::env::current_dir().unwrap().join(ZEBRA_CONFIG),
+            NodeKind::Zcashd => std::env::current_dir().unwrap().join(ZCASHD_CONFIG),
+        }
+    }
+}
+
 /// Node configuration read from the `config.toml` file.
 #[derive(Clone)]
 pub(super) struct NodeMetaData {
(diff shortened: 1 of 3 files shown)


src/setup/node.rs
```diff
@@ -67,11 +67,35 @@ impl Node {
...
 
-        Ok(Self {
+        let mut node = Self {
             config,
             meta,
             process: None,
-        })
+        };
+
+        // insert the config file cmd args
```
Mirko-von-Leipzig, line 76: I'm not too happy with the location of this code -- but my brain is too fried to think of a better place.

Happy to get suggestions.

niklaslong: I'd probably make the change in `NodeMetaData::new` since that's where it's constructed from the args.

src/setup/node.rs
```diff
@@ -67,11 +67,35 @@ impl Node {
...
+        match node.meta.kind {
+            NodeKind::Zebra => {
+                // Bit more convoluted since we need to insert the args before `start`, which comes last.
+                let start_arg = node
+                    .meta
+                    .start_args
+                    .pop()
```
niklaslong, line 83: You could probably make an `insert` at `i = len - 2` instead?

src/setup/node.rs
```diff
@@ -67,11 +67,35 @@ impl Node {
...
+                    .expect("Expected at least one arg for Zebra (`start`)");
+                node.meta.start_args.push("--config".into());
+                node.meta
+                    .start_args
+                    .push(node.config_filepath().into_os_string());
```
niklaslong, line 88: These two previous steps could probably be condensed (perhaps with `format!` and `OsString::from`)?

niklaslong (changes requested):
This is a great idea—left a few suggestions.

src/setup/node.rs
Mirko-von-Leipzig: I was worried that having it twice in a row might be confusing (since its two separate args). But I think yours is still better.

src/setup/node.rs
Mirko-von-Leipzig: We need access to `Node::config_filepath()` for the actual location of the generated file. We could move that down a layer as well?

src/setup/node.rs
Mirko-von-Leipzig: For some reason combining the two into one, results in extra quotes around just that pair of args which breaks the command.

So instead of
```
`arg1 arg2 arg3 arg4 start`
```
we get

```
`arg1 arg2 `arg3 arg4` start`
```

src/setup/node.rs
niklaslong: Works for me 👍

src/setup/config.rs
```diff
@@ -91,6 +105,21 @@ impl NodeMetaData {
         let mut start_args = args_from(&config_file.start_command);
         let start_command = start_args.remove(0);
 
+        // insert the config file path into start args
+        let config_path = config_file.kind.config_filepath();
+        match config_file.kind {
+            NodeKind::Zebra => {
+                // Zebra's final arg must be `start`, so we insert the actual args before it.
+                let n = start_args.len();
+                assert!(n > 1, "Expected at least one arg for Zebra (`start`)");
```
niklaslong, line 114: We might actually want to return a result here instead of panicking for future proofing? (A custom runner would likely need `Result`s to function properly.

niklaslong (approved):

PR #86, opened 2021-07-05
readme: add prerequesite and logging, improve config

Mirko-von-Leipzig:
I'm not certain if this would be enough information for a novice -- but it might be enough for someone familiar with Zebra / ZCashd?

I've also removed the config field for setting the config file in anticipation of #85 getting merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,25 +55,34 @@ Similarly to `zcashd`, configuration is not necessary since Ziggurat generates n
 
 ## Configuration
 
-Ziggurat is configured with a `config.toml` file in the root.
+Ziggurat is configured via a `config.toml` file in the root directory. It must contain the following fields:
 
+- `kind`: one of `zebra` or `zcashd`.
+- `path`: absolute path in which to run the start command.
+- `start_command`: the command used to start the node
+
+We recommend using the following ZCashd config:
+```toml
+kind = "zcashd"
+path = "path/to/zcash/repo"
+start_command = "./src/zcashd -debug=1 -printtoconsole -logips=1 -dnsseed=0 -dns=0 -listenonion=0"
+# debug=1           enables debug logging
+# logips=1          adds connection IP spans to the logs
+# printtoconsole    logs to stdout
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,6 +92,20 @@ Ziggurat's documentation can be built with `cargo doc --no-deps --open`.
 
 Ziggurat currently uses rust's standard test runner, a simple `cargo test -- --test-threads=1` should suffice. We use the single threaded executor as spinning up multiple test nodes isn't currently supported.
 
+### Logging
+
+By default, the node's `stdout` and `stderr` output will not be visible. One can enable this (in the code) on a per test basis by inserting a call to `log_to_stdout(true)` before starting the node. This can be used to see the node's logs on screen if appropriately configured.
```
niklaslong, line 97: Would be great to mention `enable_tracing` for synthetic nodes too?

niklaslong (commented):
Nice! I think this is detailed enough, especially as most users of the project will be devs familiar with either zebrad, zcashd or both.

niklaslong:
Clippy breakage may be due to https://github.com/rust-lang/rust-clippy/issues/7422.

README.md
Mirko-von-Leipzig: > 
> 
> Clippy breakage may be due to [rust-lang/rust-clippy#7422](https://github.com/rust-lang/rust-clippy/issues/7422).

I think it is. It broke for me locally only after `rustup update`.

Mirko-von-Leipzig:
> 
> 
> Clippy breakage may be due to [rust-lang/rust-clippy#7422](https://github.com/rust-lang/rust-clippy/issues/7422).

Yeah I think that's the cause. Only broke for me locally after `rustup update`

README.md
Mirko-von-Leipzig: Thanks, I forgot about that!

niklaslong (approved):

niklaslong:
Let's wait for #85 so we're in sync

PR #88, opened 2021-07-06
feat: add more testnet vectors

Mirko-von-Leipzig:
Adds testnet vectors 3 through 10.

Tests have been updated to use these where applicable.

Small standardization of testnet vector bytes to make the larger heights more legible.


niklaslong (commented):
Looks good—might be worth checking these in the block roundtrip tests?

niklaslong (approved):

niklaslong (approved):

PR #91, opened 2021-07-07
ref: split reject_invalid_messages into subcases

Mirko-von-Leipzig:
Instead of running a single test case with multiple messages, we now have a single case per test message.

Failure points are now more visible; previous behavior would short-circuit on first failure which meant we never saw later failures without editing the code.

Also improved diagnostics around failure.

This does run slower since we now have to spin up `N` nodes instead of 1.

```
cargo test -- --test conformance::invalid_message --test-threads=1
    Finished test [unoptimized + debuginfo] target(s) in 0.03s
     Running unittests (target/debug/deps/ziggurat-ec25afa3121de411)

running 7 tests
test tests::conformance::invalid_message::bloom_filter_add ... FAILED
test tests::conformance::invalid_message::bloom_filter_clear ... FAILED
test tests::conformance::invalid_message::bloom_filter_load ... FAILED
test tests::conformance::invalid_message::mixed_inventory ... FAILED
test tests::conformance::invalid_message::multi_block_inventory ... FAILED
test tests::conformance::invalid_message::verack_post_handshake ... FAILED
test tests::conformance::invalid_message::version_post_handshake ... ok

failures:

---- tests::conformance::invalid_message::bloom_filter_add stdout ----
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "Incorrect rejection ccode: Malformed instead of Obsolete" }', src/tests/conformance/invalid_message.rs:56:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests::conformance::invalid_message::bloom_filter_clear stdout ----
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "Message was ignored" }', src/tests/conformance/invalid_message.rs:70:10

---- tests::conformance::invalid_message::bloom_filter_load stdout ----
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "Incorrect rejection ccode: Malformed instead of Obsolete" }', src/tests/conformance/invalid_message.rs:63:10

---- tests::conformance::invalid_message::mixed_inventory stdout ----
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "Message was ignored" }', src/tests/conformance/invalid_message.rs:36:10

---- tests::conformance::invalid_message::multi_block_inventory stdout ----
[...]

PR #93, opened 2021-07-07
ref: split ignores_unsolicited_responses

Mirko-von-Leipzig:
In the same vein as #91

### DIFF
--- a/src/tests/conformance/mod.rs
+++ b/src/tests/conformance/mod.rs
@@ -1,36 +1,4 @@
-use std::io;
-
-use crate::{
-    setup::node::{Action, Node},
-    tools::synthetic_node::SyntheticNode,
-};
-
 mod handshake;
 mod invalid_message;
 mod messages;
 mod unsolicited_response;
-
-/// Creates a connected [Node] and [SyntheticNode] which have completed the full handshake.
-///
-/// The [SyntheticNode] initiated the connection.
-///
-/// The [SyntheticNode]'s MessageFilter has been configured for auto replying.
-async fn simple_handshaken_node() -> io::Result<(Node, SyntheticNode)> {
-    // Spin up a node instance.
-    let mut node = Node::new()?;
-    node.initial_action(Action::WaitForConnection)
-        .start()
-        .await?;
-
-    // Create a synthetic node.
-    let synthetic_node = SyntheticNode::builder()
-        .with_full_handshake()
-        .with_all_auto_reply()
-        .build()
-        .await?;
-
-    // Connect and initiate the handshake.
-    synthetic_node.connect(node.addr()).await?;
-
-    Ok((node, synthetic_node))
-}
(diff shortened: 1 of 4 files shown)


src/tests/conformance/mod.rs
```diff
@@ -1,2 +1,37 @@
...
+mod unsolicited_response;
+
+/// Creates a connected [Node] and [SyntheticNode] which have completed the full handshake.
+///
+/// The [SyntheticNode] initiated the connection.
+///
+/// The [SyntheticNode]'s MessageFilter has been configured for auto replying.
+async fn simple_handshaken_node() -> io::Result<(Node, SyntheticNode)> {
+    // Spin up a node instance.
+    let mut node = Node::new().unwrap();
```
niklaslong, line 19:
```suggestion
    let mut node = Node::new()?;
```

src/tests/conformance/unsolicited_response.rs
```diff
@@ -0,0 +1,116 @@
+///! Contains test cases which cover ZG-CONFORMANCE-009.
+///!
+///! The node should ignore the following unsolicited messages:
+///!
+///!  Reject, NotFound, Pong, Tx, Block, Header, Addr
```
niklaslong, line 5:
```suggestion
///!  Reject, NotFound, Pong, Tx, Block, Header, Addr

```

src/tests/conformance/unsolicited_response.rs
```diff
@@ -0,0 +1,116 @@
...
+        .unwrap();
+}
+
+async fn run_test_case(message: Message) -> io::Result<()> {
+    let (mut node, mut synthetic_node) = simple_handshaken_node().await?;
+
+    synthetic_node
+        .send_direct_message(node.addr(), message)
+        .await
+        .unwrap();
```
niklaslong, line 79:
```suggestion
        .await?;
```

src/tests/conformance/mod.rs
```diff
@@ -1,3 +1,36 @@
...
 mod invalid_message;
 mod messages;
+mod unsolicited_response;
+
+/// Creates a connected [Node] and [SyntheticNode] which have completed the full handshake.
+///
+/// The [SyntheticNode] initiated the connection.
+///
+/// The [SyntheticNode]'s MessageFilter has been configured for auto replying.
+async fn simple_handshaken_node() -> io::Result<(Node, SyntheticNode)> {
```
niklaslong, line 18:
```suggestion
async fn simple_handshaken_node_and_synth_node() -> io::Result<(Node, SyntheticNode)> {
```

niklaslong: The downside to this is we loose the explictness of the setup in the test cases 🤔

niklaslong (approved):

src/tests/conformance/mod.rs
Mirko-von-Leipzig: I don't mind removing this again.

Naming could definitely be better. Some other options:

1. `setup_basic_handshaken_connection`
2. `node_with_single_handshaken_connection`

niklaslong: I think it might be best to keep the cases flexible albeit a little less DRY for now since the suite is still fairly young and will likely change in future?

Mirko-von-Leipzig: Sure, I can just move it up one notch back into the same file at least.

niklaslong (approved):

PR #96, opened 2021-07-08
refactor: split basic query tests

Mirko-von-Leipzig:
Includes adding a custom error type for `ping_pong_timeout` which improves the ability of tests to check the node's response to a prior query.


src/tools/synthetic_node.rs
```diff
@@ -26,6 +26,64 @@ use std::{
     time::Duration,
 };
 
+/// An [`Error`](std::error::Error) type for [`SyntheticNode::ping_pong_timeout()`]
```
niklaslong, line 29:
```suggestion
/// An [`Error`](std::error::Error) type for [`SyntheticNode::ping_pong_timeout`]
```

src/tools/synthetic_node.rs
```diff
@@ -26,6 +26,64 @@ use std::{
...
+pub enum PingPongError {
+    /// The connection was aborted during the [Ping](Message::Ping)-[Pong](Message::Pong) exchange.
```
niklaslong, line 31:
```suggestion
    /// The connection was aborted during the [`Ping`](Message::Ping)-[`Pong`](Message::Pong) exchange.
```

niklaslong: The same applies to the docs below.

src/tools/synthetic_node.rs
```diff
@@ -241,10 +299,17 @@ impl SyntheticNode {
 
     /// Sends [`Ping`], and expects [`Pong`] with a matching [`Nonce`] in reply.
     ///
-    /// Uses polling to check that connection is still alive. Errors if:
+    /// Uses polling to check that connection is still alive. [Errors](PingPongError) if:
```
niklaslong, line 302:
```suggestion
    /// Uses polling to check that connection is still alive. Returns an [`PingPongError`](PingPongError) if:
```

src/tools/synthetic_node.rs
```diff
@@ -241,10 +299,17 @@ impl SyntheticNode {
...
     /// - a non-[`Pong`] message is received
+    /// - a [`Pong`] with a non-matching [`Nonce`] is receives
     /// - the timeout expires
     /// - the connection breaks
+    /// - an [io::Error] occurs
```
niklaslong, line 307:
```suggestion
    /// - an [`io::Error`] occurs
```

niklaslong (commented):
This is great! I left a few comments regarding documentation but otherwise this should be good to go. 👍

niklaslong (approved):

PR #97, opened 2021-07-09
feat: remove old query tests

Mirko-von-Leipzig:
Forgot to remove these when splitting them up.


Mirko-von-Leipzig:
Nevermind -- I forgot to update master, they're already gone.

PR #98, opened 2021-07-09
refactor: invalid message disconnection tests

Mirko-von-Leipzig:
- moved the peering related tests
- moved and split the disconnect for trivial cases tests
- added disconnect test for `Addr` with missing timestamp
- added disconnect test for `Ping` timeout (`#[ignore]`)
- added `Message::short_string()` - useful to not spam the terminal full of debug noise

The two added tests could definitely use more scrutiny.


src/tests/conformance/invalid_message/disconnect.rs
```diff
@@ -0,0 +1,303 @@
...
+    tools::{
+        message_filter::{Filter, MessageFilter},
+        synthetic_node::{PingPongError, SyntheticNode},
+    },
+};
+
+const DC_TIMEOUT: Duration = Duration::from_secs(1);
+
+#[tokio::test]
+#[ignore = "ZCashd ping timeout is 20min"]
```
Mirko-von-Leipzig, line 40: I don't know if this is good - since it will now clash with our internal protocol tests (I think?).

I do want to include the test but I don't quite know how else.

src/tests/conformance/invalid_message/disconnect.rs
```diff
@@ -0,0 +1,303 @@
...
+/// This is used by the [`addr_without_timestamp()`] test. [`Message::Addr`]
+/// cannot be used for this, as it does not support encoding without a timestamp (as
+/// this is obsolete behaviour).
+struct AddrWithoutTimestamp(Vec<NetAddrWithoutTimestamp>);
+impl AddrWithoutTimestamp {
+    fn new() -> Self {
+        Self(vec![NetAddrWithoutTimestamp::new()])
+    }
+
+    fn encode(&self) -> io::Result<Vec<u8>> {
```
Mirko-von-Leipzig, line 240: I think this is correct - but I did spend a lot of time looking and re-doing this, so I might have blind-sight. Additional scrutiny would be appreciated.

src/tests/conformance/invalid_message/disconnect.rs
niklaslong: `--lib` wil run our entire test suite (since the `tests` are in `lib`). Feature flags might be an option. Another possibility is to have two crates: `core` and `suite`, though I'm hesitant to make structural changes of that magnitude this late in the grant. This would best be left for later. 

Simply removing `#[tokio::test]` and adding a comment is probably best, especially since no one is ever actually going to run this test case. For that reason I'm actually partial to removing the case entirely. (As long as the reason for doing so is documented somewhere obvious.)

src/tests/conformance/invalid_message/disconnect.rs
```diff
@@ -0,0 +1,303 @@
...
+    // zebra:  fail (replies with Reject(Malformed))
+
+    // Encode a custom type which mimics Message::Addr but without the timestamp.
+    let bytes = AddrWithoutTimestamp::new().encode().unwrap();
+
+    run_test_case_bytes(bytes).await.unwrap();
+}
+
+/// Mimics the encoding of [`NetworkAddr`] but excludes the timestamp field.
+struct NetAddrWithoutTimestamp(NetworkAddr);
```
niklaslong, line 206: For consistency with `NetworkAddr`:
```suggestion
struct NetworkAddrWithoutTimestamp(NetworkAddr);
```

src/tests/conformance/invalid_message/disconnect.rs
niklaslong: Another option for this is to use the `NetworkAddr` implementation and remove the range of bytes corresponding to the timestamp with e.g. `Vec::drain`.

niklaslong (commented):
Left a few suggestions, great work on this! 👍

src/tests/conformance/invalid_message/disconnect.rs
Mirko-von-Leipzig: I think your idea is a lot cleaner. Will change to that.

src/tests/conformance/invalid_message/disconnect.rs
Mirko-von-Leipzig: Okay, will add a comment and remove it.

src/protocol/message/mod.rs
```diff
@@ -192,6 +192,32 @@ impl Message {
 
         Ok(message)
     }
+
+    /// Returns a short string describing at least the message type.
+    /// Should be suitable for display purposes.
+    pub fn short_string(&self) -> String {
```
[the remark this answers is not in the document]
Mirko-von-Leipzig: I'm happy with `Display`

src/tests/conformance/invalid_message/disconnect.rs
```diff
@@ -193,63 +115,34 @@ async fn inv_with_mixed_types() {
...
+    let message = Message::Addr(Addr::new(vec![NetworkAddr::new(SocketAddr::new(
+        IpAddr::V4(Ipv4Addr::new(1, 2, 3, 4)),
+        10,
+    ))]));
+    let mut payload = Vec::new();
+    message.encode(&mut payload).unwrap();
+
+    // Remove the timestamp bytes.
+    //
+    // First byte is a VarInt=1 containing the length of our NetworkAddr vector (1).
```
niklaslong, line 131: Note: the number of bytes will change if the test is modified to use more addresses, perhaps something to make _extra_ clear here, as this is easy to overlook.

niklaslong (approved):

src/tests/conformance/invalid_message/disconnect.rs
Mirko-von-Leipzig: I've opted to make the length calculation explicit. Let me know if you prefer this; or rather a comment?

The change does require adding a `pub` constructor for `VarInt`.

niklaslong (approved):

PR #100, opened 2021-07-15
ref: split up get_data tests

Mirko-von-Leipzig:


src/tests/conformance/get_data.rs
```diff
@@ -1,129 +1,237 @@
...
+            .iter()
+            .map(|&block| Message::Block(Box::new(block.clone())))
+            .collect::<Vec<_>>();
+        let expected_non_existant = Message::NotFound(Inv::new(non_existant_inv.clone()));
+
+        let mixed_inv = blocks
+            .iter()
+            .map(|block| block.inv_hash())
+            .zip(non_existant_inv)
+            .map(|(a, b)| [a, b])
```
Mirko-von-Leipzig, line 174: I feel like one should be able to `flat_map` here, but I could not please the borrow checker.

niklaslong: The following should work fine?
```
        let mixed_inv = blocks
            .iter()
            .map(|block| block.inv_hash())
            .zip(non_existant_inv)
            .flat_map(|(a, b)| [a, b])
            .collect();
```

Mirko-von-Leipzig: Ah I was trying 
`.flat_map(|(a, b)| [a, b].into_iter())`
Thanks!

src/tests/conformance/get_data.rs
```diff
@@ -1,129 +1,235 @@
...
-        Message::NotFound(non_existant_inv.clone()),
-        Message::Block(Box::new(Block::testnet_1())),
-        Message::Block(Box::new(Block::testnet_2())),
-        Message::NotFound(non_existant_inv),
-    ];
+        .send_direct_message(node.addr(), query)
+        .await?;
 
+    // Send a Ping - once we receive the matching Pong we know our query has been fully processed.
+    let nonce = Nonce::default();
```
niklaslong, line 215: Did we get rid of the ping/pong assertion helper on `SyntheticNode`?

Mirko-von-Leipzig: No its still there - but it only receives a single message before returning.

`GetData` elicits a `Block` message for every known hash in the request. So in this case the looping receive is needed.

One could change / extend a multi version on the `SyntheticNode`, but so far this is the only use case for it. So I opted to just leave it here.

niklaslong (approved):

PR #101, opened 2021-07-16
refactor: get_headers conformance tests

Mirko-von-Leipzig:
Last of the conformance test refactoring.


src/tests/conformance/query/get_header.rs
```diff
@@ -1,139 +1,344 @@
...
-        .await
-        .unwrap();
+/// A wrapper around [`run_test_query`] which maps its output to [`Response`].
+async fn run_test_case(query: GetHeaders) -> io::Result<Response> {
+    let mut reply = run_test_query(query.0).await?;
 
-    let (_, headers) = synthetic_node.recv_message_timeout(TIMEOUT).await.unwrap();
-    let headers = assert_matches!(headers, Message::Headers(headers) => headers);
-    assert_eq!(headers.headers, expected[2..], "test for forked Headers");
+    let response = if reply.is_empty() {
```
niklaslong, line 330: nit/suggestion: `match` on `reply.len()`

niklaslong (approved):
Nice! 🙌 

Would `basic_query.rs` fit in this new `query` dir?

Mirko-von-Leipzig:
> 
> 
> Nice! 🙌
> 
> Would `basic_query.rs` fit in this new `query` dir?

I've moved the file in, and fixed the `match len` nit (x2).

I also renamed `get_header.rs` to be plural - this matches the message struct etc.

PR #104, opened 2021-07-23
fix: only clear testnet cache

niklaslong:
Removes only testnet related cached directories when cleaning up after tests. Unfortunately a custom directory can't be specified at the moment (Ziggurat runs with testnet blocks), any testnet state will be cleared by Ziggurat.


Mirko-von-Leipzig (approved):
LGTM

Should we consider a note in the README (that we will nuke the testnet cache)?

PR #105, opened 2021-07-23
feat: move configuration and node caches to `~/.ziggurat` directory

### LINKED ISSUE #103: If you don't specify a -datadir argument to zcashd and run the tool, it seems to delete the contents of any ~/.zcash directory that exists (potential loss of funds issue)

niklaslong:
Supersedes #104 and closes #103.

### DIFF
--- a/src/setup/config.rs
+++ b/src/setup/config.rs
@@ -2,9 +2,9 @@ use serde::{Deserialize, Serialize};
 
 use std::{
     collections::HashSet,
-    env,
     ffi::OsString,
     fs, io,
+    io::{Error, ErrorKind},
     net::{IpAddr, Ipv4Addr, SocketAddr},
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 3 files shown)


src/setup/config.rs
```diff
@@ -118,10 +135,10 @@ impl NodeMetaData {
                     ));
                 }
                 start_args.insert(n_args - 1, "--config".into());
-                start_args.insert(n_args, config_path.into_os_string());
+                start_args.insert(n_args, config_file_path.into_os_string());
             }
             NodeKind::Zcashd => {
-                start_args.push(format!("-conf={}", config_path.to_str().unwrap()).into());
+                start_args.push(format!("-datadir={}", config_path.to_str().unwrap()).into());
```
Mirko-von-Leipzig, line 141: What happens if the user adds his own `-datadir` arg? Which one wins?

Mirko-von-Leipzig (approved):
LGTM

src/setup/config.rs
niklaslong: Any user defined `-datadir` arg will get ovveridden by Ziggurat's.

daira: This should be documented in the readme.

PR #108, opened 2021-07-30
Update pea2pea and metrics deps

niklaslong:
Updates the deps. The new pea2pea version allows for the removal of some async functions in favour of synchronous versions.


Mirko-von-Leipzig (commented):
LGTM!

Did you re-run some tests as a sanity check?

Mirko-von-Leipzig (approved):

niklaslong:
I have, this runs as expected with both `zebrad` and `zcashd` nodes.

PR #112, opened 2022-03-25
Update pea2pea to 0.35

ljedrz:
This PR updates `pea2pea` to `0.35`. The library has changed significantly since then, hence the rather large diff.

TODO (either as extensions to this PR or as follow-ups):

- [x] fix the corrupting test functions (currently commented out and replaced with the `todo` macro
- [ ] check buffer lengths in `decode` functions or wrap the "outermost" call in `catch_unwind`


ljedrz:
Fixed the corrupting functions.

I propose to take care of the other TODO as a follow-up, as the synthetic node currently trusts its inputs (which is perfectly fine for test purposes, but can be easily improved to not panic).

src/protocol/message/mod.rs
```diff
@@ -89,82 +98,89 @@ pub enum Message {
     FilterClear,
 }
 
+macro_rules! enode_with_header_prefix {
```
niklaslong, line 101: Nice! 👍

niklaslong: Almost missed it but:
```suggestion
macro_rules! encode_with_header_prefix {
```

niklaslong (approved):
LGTM, minus that one typo 👍

ljedrz:
Typo fixed :ok_hand:.

PR #113, opened 2022-04-06
Feat: Nu5 Tx (v5) updates

niklaslong:
See: https://zips.z.cash/protocol/protocol.pdf#networkupgrades

### DIFF
--- a/src/protocol/payload/tx.rs
+++ b/src/protocol/payload/tx.rs
@@ -18,8 +18,7 @@ pub enum Tx {
     V2(TxV2),
     V3(TxV3),
     V4(TxV4),
-    // Not yet stabilised.
-    V5,
+    V5(TxV5),
 }
 
 impl Tx {
@@ -66,7 +65,11 @@ impl Codec for Tx {
                 buffer.put_u32_le(4u32 | 1 << 31);
                 tx.encode(buffer)?;
             }
-            Tx::V5 => unimplemented!(),
+            Tx::V5(tx) => {
+                // The overwintered flag IS set.
+                buffer.put_u32_le(5u32 | 1 << 31);
+                tx.encode(buffer)?;
+            }
         }
 
         Ok(())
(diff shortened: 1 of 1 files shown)


src/protocol/payload/tx.rs
```diff
@@ -421,6 +424,236 @@ impl Codec for TxV4 {
...
+        self.actions_orchard.encode(buffer)?;
+
+        if !self.actions_orchard.is_empty() {
+            buffer.put_u8(self.flags_orchard.unwrap());
+            buffer.put_i64_le(self.value_balance_orchard.unwrap());
+            buffer.put_slice(&self.anchor_orchard.unwrap());
+
+            VarInt(self.proofs_orchard.as_ref().unwrap().len()).encode(buffer)?;
+            for proof in self.proofs_orchard.as_ref().unwrap() {
+                buffer.put_u8(*proof)
```
ljedrz, line 509: `proofs_orchard` can probably be encoded "wholesale" here via `put_slice`

src/protocol/payload/tx.rs
```diff
@@ -421,6 +424,236 @@ impl Codec for TxV4 {
...
+        let (
+            flags_orchard,
+            value_balance_orchard,
+            anchor_orchard,
+            proofs_orchard,
+            auth_sigs_orchard,
+            binding_sig_orchard,
+        ) = if !actions_orchard.is_empty() {
+            // Decode the orchard flags.
+            if bytes.remaining() < 1 {
```
ljedrz, line 585:
```suggestion
            if bytes.remaining() == 0 {
```

src/protocol/payload/tx.rs
```diff
@@ -785,4 +1125,37 @@ mod tests {
 
         assert_eq!(tx_v4, Tx::decode(&mut Cursor::new(&bytes)).unwrap());
     }
+
+    #[test]
+    #[ignore]
```
ljedrz, line 1130: why do we ignore this? also, it would be best to test with a non-empty tx, preferably from some original test vector

ljedrz (approved):
LGTM, left a few comments.

src/protocol/payload/tx.rs
niklaslong: We ignored all the unit tests so that we can run Ziggurat with just `cargo t`, though perhaps a feature flag would be more appropriate in the long run.

niklaslong: > preferably from some original test vector
Agreed, I'll follow up with this if I can find some.

PR #114, opened 2022-04-06
Feat: update the network protocol version number to Nu5

### LINKED ISSUE #110: Document protocol version
I had a hard time trying to figure out why everything was failing in the recent version of Zebra until i found that was the protocol version that needs to be updated: https://github.com/eqlabs/ziggurat/blob/master/src/protocol/payload/mod.rs#L65

I think it will be good to document this in the README and/or move the constant to https://github.com/eqlabs/ziggurat/blob/master/src/setup/config.rs

niklaslong:
Changes the version number to `170_015` and introduces a constant holding the value. This has also been documented in the README. Closes #110.

Branch based on #113, only the last commit is new.


ljedrz (approved):
LGTM :+1:

PR #115, opened 2022-04-08
Feat: Nu5 transition test vectors (testnet)

niklaslong:
This PR adds the test vectors for the nu5 block transitions (testnet) and introduces some encoding/decoding tests.

The blocks were borrowed from [Zebra](https://github.com/ZcashFoundation/zebra/tree/main/zebra-test/src/vectors).


src/protocol/payload/block.rs
```diff
@@ -591,6 +591,54 @@ mod tests {
         assert_eq!(block_bytes, buffer);
     }
 
+    #[test]
+    #[ignore]
```
ljedrz, line 595: why are these ignored? doesn't seem like something slow

ljedrz (approved):
LGTM :ok_hand:

src/protocol/payload/block.rs
niklaslong: All unit tests are ignored currently, so we can distinguish them from the Ziggurat suite and run the latter with `cargo test`.

PR #118, opened 2022-04-12
Add a fuzz test for the synthetic node impl

ljedrz:
This PR adds a `decoder` fuzz target that can be used with `cargo-fuzz` (`cargo +nightly fuzz run decoder`) and a few patches that fix issues that I've found with it, now that the synthetic node can be used outside of tests.

Cc @tuommaki

### DIFF
--- a/src/protocol/payload/addr.rs
+++ b/src/protocol/payload/addr.rs
@@ -86,6 +86,10 @@ impl NetworkAddr {
     pub(super) fn decode_without_timestamp<B: Buf>(bytes: &mut B) -> io::Result<Self> {
         let services = u64::from_le_bytes(read_n_bytes(bytes)?);
 
+        if bytes.remaining() < 16 {
+            return Err(io::ErrorKind::InvalidData.into());
+        }
+
         let mut octets = [0u8; 16];
         bytes.copy_to_slice(&mut octets);
         let v6_addr = Ipv6Addr::from(octets);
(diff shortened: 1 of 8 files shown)


src/protocol/payload/codec.rs
```diff
@@ -32,8 +32,6 @@ impl<T: Codec> Codec for Vec<T> {
         Self: Sized,
     {
         let length = *VarInt::decode(bytes)?;
-        (0..length)
-            .map(|_| T::decode(bytes))
-            .collect::<io::Result<Self>>()
+        (0..length).map(|_| T::decode(bytes)).collect()
```
ljedrz, line 35: this one is just a drive-by cleanup

src/protocol/payload/addr.rs
```diff
@@ -86,6 +86,10 @@ impl NetworkAddr {
     pub(super) fn decode_without_timestamp<B: Buf>(bytes: &mut B) -> io::Result<Self> {
         let services = u64::from_le_bytes(read_n_bytes(bytes)?);
 
+        if bytes.remaining() < 16 {
```
niklaslong, line 89: We could consider using `read_n_bytes` everywhere instead, since it checks for the length. Not sure if it would be slower than the built in `get_*` methods.

niklaslong (approved):
LGTM 👍

src/protocol/payload/addr.rs
ljedrz: yeah, I had the same thoughts; we might want to align this at one point in time, but I'm ok with both options

PR #122, opened 2022-04-27
Timeout constant unification for conformance tests

niklaslong:
This starts unifying the various timeout constants in the context of conformance tests. In future the same should be done for the performance and resistance tests. They should likely also be made configurable in Ziggurat's configuration file.

### DIFF
--- a/src/tools/mod.rs
+++ b/src/tools/mod.rs
@@ -9,7 +9,7 @@ use std::time::Duration;
 
 /// Default timeout for connection operations in seconds.
 /// TODO: move to config file.
-pub const TIMEOUT: Duration = Duration::from_secs(10);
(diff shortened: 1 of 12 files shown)


src/tools/mod.rs
```diff
@@ -7,8 +7,11 @@ pub mod synthetic_node;
 
 use std::time::Duration;
 
-/// Default timeout for connection reads in seconds.
+/// Default timeout for connection operations in seconds.
```
ljedrz, line 10: the `TIMEOUT` name is a bit too broad IMO, same with `connection operations`; we should probably have one for starting a full node and, if we need another one for some connection-specific thing, have a separate one (with a more explicit name too) for it as well

ljedrz (approved):
LGTM :ok_hand:

PR #124, opened 2022-05-04
Introduce symbolic language for test specification

niklaslong:
Uses symbolic language to better describe the protocol specifics of conformance and resistance tests. The two performance tests have been left out as their specification is vague (perhaps this should be improved as well).


ljedrz (approved):
LGTM :ok_hand:

PR #125, opened 2022-05-10
Introduce a network-based crawler

ljedrz:
It's already operational (checked against the testnet with the updated protocol version), but it'll likely require further tweaks and enhancements; it will also need to be moved from a test to a binary.

In addition, `pea2pea` is updated to `0.37` and a timestamp-handling fix is applied for `addr` messages.


src/protocol/payload/addr.rs
```diff
@@ -124,7 +124,7 @@ impl Codec for NetworkAddr {
     }
 
     fn decode<B: Buf>(bytes: &mut B) -> io::Result<Self> {
-        let timestamp = read_timestamp(bytes)?;
+        let timestamp = read_short_timestamp(bytes)?;
```
niklaslong, line 127: This protocol change occured without any change to the message command string in the header?

src/protocol/payload/mod.rs
```diff
@@ -241,3 +241,10 @@ pub fn read_timestamp<B: Buf>(bytes: &mut B) -> io::Result<OffsetDateTime> {
     OffsetDateTime::from_unix_timestamp(timestamp_i64)
         .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "Bad UTC timestamp"))
 }
+
+/// Reads a timestamp from the bytes.
```
niklaslong, line 245: Suggestion:
```suggestion
/// Reads a timestamp encoded as 4 bytes.
```

Ditto for the 8-byte variant.

src/tools/crawler.rs
```diff
@@ -0,0 +1,380 @@
...
+    pub last_connected: Option<Instant>,
+    /// The time it took to complete a connection.
+    pub handshake_time: Option<Duration>,
+    /// The number of subsequent connection errors.
+    connection_failures: u8,
+}
+
+/// A connection found in the network.
+#[derive(Debug, Eq, Copy, Clone)]
+pub struct KnownConnection {
```
niklaslong, line 43: Note: this could easily be replaced by `spectre` in future.

src/tools/crawler.rs
```diff
@@ -0,0 +1,380 @@
...
+    known_network: Arc<KnownNetwork>,
+}
+
+impl Pea2Pea for Crawler {
+    fn node(&self) -> &Pea2PeaNode {
+        &self.node
+    }
+}
+
+impl Crawler {
```
niklaslong, line 155: We'll probably want to build a nice cli for the crawler, in future.

niklaslong (approved):
Left a few comments but otherwise this looks good to me 👍

src/protocol/payload/addr.rs
ljedrz: I think `addr` message hasn't worked in tests before (possibly due to a long reply time?), so it's the same message as before - it might be the only one with a timestamp this short.

PR #129, opened 2022-06-22
Disable Crawler's listener ip

zeapoz:
Removes `listener_ip` from the config of `Crawler` so we don't collect duplicate nodes, for now it also provides a fake `SocketAddr` during the handshake.


ljedrz (approved):
LGTM

PR #133, opened 2022-06-29
Moved crawler to own directory

zeapoz:
Moves crawler into a new directory at `/src/tools/crawler/` and separates it between `bin.rs` and `network.rs`.


ljedrz (approved):
LGTM :+1:

PR #134, opened 2022-06-29
ref: added magic for testnet and mainnet

zeapoz:
Adds magic for testnet and mainnet, also bumps protocol version to `170_100`.


ljedrz (approved):
LGTM :+1:

PR #141, opened 2022-07-01
Summary tweaking

zeapoz:
`NetworkSummary` now takes a reference to Crawler instead and has a new field for number of nodes that identifies themselves with a `Version`.


ljedrz (approved):
LGTM :+1:

PR #142, opened 2022-07-04
Reuse existing metrics graph

zeapoz:
Now creates `Graph` only once and updates it progressively.

### DIFF
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -58,7 +58,7 @@ async fn main() {
     // Create the crawler with the given listener address.
     let crawler = Crawler::new().await;
 
-    let mut network_metrics = NetworkMetrics::new();
+    let mut network_metrics = NetworkMetrics::default();
 
     crawler.enable_handshake().await;
     crawler.enable_reading().await;
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -14,12 +14,12 @@ use ziggurat::{protocol::message::Message, wait_until};
 use crate::{
     protocol::{Crawler, MAIN_LOOP_INTERVAL, NUM_CONN_ATTEMPTS_PERIODIC},
     network::KnownNode,
-    summary::NetworkSummary,
+    metrics::NetworkMetrics,
 };
 
 mod protocol;
 mod network;
-mod summary;
+mod metrics;
 
 #[derive(Parser)]
 #[clap(author, version, about, long_about = None)]
@@ -57,6 +57,8 @@ async fn main() {
 
     // Create the crawler with the given listener address.
     let crawler = Crawler::new().await;
+
+    let mut network_metrics = NetworkMetrics::new();
 
     crawler.enable_handshake().await;
     crawler.enable_reading().await;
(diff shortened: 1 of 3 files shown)


src/tools/crawler/metrics.rs
```diff
@@ -4,13 +4,54 @@ use std::{
...
+        let (drained, connections): (Vec<_>, Vec<_>) = crawler.known_network.connections()
+            .into_iter()
+            .partition(|&conn| {
+                conn.last_seen.elapsed().as_secs() > LAST_SEEN_CUTOFF
+            });
+
+        // Add new connections.
+        for conn in connections {
+            let edge = Edge::new(conn.a, conn.b);
+            if !(self.graph.contains(&edge)) {
```
ljedrz, line 38: we shouldn't do a lookup here, as the `Edges` are identical (so replacing an existing one with a new one does nothing); it's faster to just do `self.graph.insert(edge);` and not care

src/tools/crawler/metrics.rs
```diff
@@ -28,7 +69,7 @@ pub struct NetworkSummary {
 
 impl NetworkSummary {
     /// Constructs a new NetworkSummary from given nodes.
-    pub fn new(crawler: &Crawler) -> NetworkSummary {
+    pub fn new(crawler: &Crawler, graph: &mut Graph<SocketAddr>) -> NetworkSummary {
```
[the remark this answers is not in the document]
ljedrz: it actually _is_ needed :detective:

ljedrz (commented):
LGTM in general, left some comments

src/tools/crawler/metrics.rs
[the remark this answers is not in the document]
niklaslong: It currently is but as discussed, this is something that should be changed in `spectre`.

src/tools/crawler/network.rs
```diff
@@ -125,9 +125,7 @@ impl KnownNetwork {
 
         let mut nodes = self.nodes.write();
         for addr in prospect_nodes {
-            if !nodes.contains_key(&addr) {
-                nodes.insert(addr, KnownNode::default());
-            }
+            nodes.entry(addr).or_insert_with(KnownNode::default);
```
ljedrz, line 128:
```suggestion
            nodes.entry(addr).or_default();
```
wouldn't just this one work here :detective:?

src/tools/crawler/metrics.rs
```diff
@@ -4,13 +4,45 @@ use std::{
...
+pub struct NetworkMetrics {
+    graph: Graph<SocketAddr>,
+}
+
+impl NetworkMetrics {
+    /// Constructs a new NetworkMetrics with an empty graph.
+    pub fn new() -> NetworkMetrics {
+        NetworkMetrics::default()
+    }
+
```
ljedrz, line 28:
```suggestion
```

src/tools/crawler/main.rs
```diff
@@ -58,6 +58,8 @@ async fn main() {
     // Create the crawler with the given listener address.
     let crawler = Crawler::new().await;
 
+    let mut network_metrics = NetworkMetrics::new();
```
ljedrz, line 61:
```suggestion
    let mut network_metrics = NetworkMetrics::default();
```

ljedrz (approved):
LGTM :ok_hand:

PR #143, opened 2022-07-05
doc: update zebra install instructions

niklaslong:
Includes parameter download as an install step.


ljedrz (approved):
looks reasonable :ok_hand:

PR #148, opened 2022-07-07
Only send `GetAddr` request once to seed nodes initially

zeapoz:
Instead of sending requests each iteration, crawler now waits for a response and then starts the main crawling loop. This PR also improves the crawler's general error handling.

### DIFF
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -21,6 +21,9 @@ mod metrics;
 mod network;
 mod protocol;
 
+const SEED_WAIT_LOOP_INTERVAL_MS: u64 = 500;
+const SEED_RESPONSE_TIMEOUT_MS: u64 = 120_000;
+
 #[derive(Parser)]
 #[clap(author, version, about, long_about = None)]
 struct Args {
@@ -64,10 +67,11 @@ async fn main() {
     crawler.enable_reading().await;
     crawler.enable_writing().await;
 
-    for addr in args.seed_addrs {
+    for addr in &args.seed_addrs {
         let crawler_clone = crawler.clone();
+        let addr = addr.parse().unwrap();
+
         tokio::spawn(async move {
-            let addr = addr.parse().unwrap();
             crawler_clone
                 .known_network
                 .nodes
@@ -81,10 +85,12 @@ async fn main() {
         });
     }
 
-    // Wait for the connection to be complete.
-    wait_until!(Duration::from_secs(3), crawler.node().num_connected() >= 1);
-
-    sleep(Duration::from_secs(1)).await;
+    // Wait for one of the seed nodes to respond with a list of addrs.
+    wait_until!(
+        Duration::from_millis(SEED_RESPONSE_TIMEOUT_MS),
+        crawler.known_network.nodes().len() > args.seed_addrs.len(),
+        Duration::from_millis(SEED_WAIT_LOOP_INTERVAL_MS)
+    );
 
     tokio::spawn(async move {
         loop {
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -7,7 +7,7 @@ use pea2pea::{
 };
 use rand::prelude::IteratorRandom;
 use tokio::time::sleep;
-use tracing::info;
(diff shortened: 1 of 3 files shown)


src/tools/crawler/main.rs
```diff
@@ -108,15 +109,18 @@ async fn main() {
...
                 // Update graph, then create a summary and log it to a file.
                 network_metrics.update_graph(&crawler);
                 let network_summary = network_metrics.request_summary(&crawler);
 
                 info!("{}", network_summary);
-                network_summary.log_to_file().unwrap();
+                network_summary.log_to_file().unwrap_or_else(|e| {
+                    error!(parent: crawler.node().span(), "Couldn't write summary to file: {}", e);
+                    ()
+                });
```
ljedrz, line 123: this would be expressed better as `if let Err(e) = network_summary.log_to_file() { error!(...) }`

src/tools/crawler/main.rs
```diff
@@ -81,8 +82,8 @@ async fn main() {
         });
     }
 
-    // Wait for the connection to be complete.
-    wait_until!(Duration::from_secs(3), crawler.node().num_connected() >= 1);
+    // Wait for one of the seed nodes to respond with a list of addrs.
+    wait_until!(Duration::from_secs(120), crawler.known_network.nodes().len() > args.seed_addrs.len());
```
ljedrz, line 86: since we can potentially need to wait for a long time, it's probably best that we alter the check interval for `wait_until` between `#[cfg(test)]` (the current value) and `#[cfg(not(test))]` (probably sth like 200ms)

src/tools/crawler/main.rs
```diff
@@ -108,15 +116,19 @@ async fn main() {
                 }
             }
 
-            crawler.send_broadcast(Message::GetAddr).unwrap();
+            crawler
+                .send_broadcast(Message::GetAddr)
+                .expect("writing hasn't been enabled");
```
ljedrz, line 121: this can remain an `unwrap()`; we know that it's enabled

src/tools/crawler/main.rs
```diff
@@ -108,15 +116,19 @@ async fn main() {
...
             if crawler.known_network.num_connections() > 0 {
                 // Update graph, then create a summary and log it to a file.
                 network_metrics.update_graph(&crawler);
                 let network_summary = network_metrics.request_summary(&crawler);
 
                 info!("{}", network_summary);
-                network_summary.log_to_file().unwrap();
+                if let Err(e) = network_summary.log_to_file() {
+                    error!(parent: crawler.node().span(), "Couldn't write summary to file: {}", e);
+                };
```
ljedrz, line 131:
```suggestion
                }
```

src/tools/crawler/main.rs
```diff
@@ -21,6 +21,13 @@ mod metrics;
 mod network;
 mod protocol;
 
+#[cfg(test)]
+const ADDRS_FROM_SEED_SLEEP_DURATION: u64 = 10;
+#[cfg(not(test))]
+const ADDRS_FROM_SEED_SLEEP_DURATION: u64 = 200;
+
+const ADDRS_FROM_SEED_TIMEOUT: u64 = 120_000;
```
ljedrz, line 29:
```suggestion
const SEED_RESPONSE_TIMEOUT: u64 = 120_000;
```

src/tools/crawler/main.rs
```diff
@@ -21,6 +21,10 @@ mod metrics;
 mod network;
 mod protocol;
 
+#[cfg(not(test))]
+const LOOP_INTERVAL_MS: u64 = 200;
```
ljedrz, line 25:
```suggestion
#[cfg(not(test))]
const SEED_LOOP_INTERVAL_MS: u64 = 200;
```

src/tools/crawler/main.rs
```diff
@@ -21,6 +21,10 @@ mod metrics;
...
+const SEED_RESPONSE_TIMOUT: u64 = 120_000;
```
ljedrz, line 26:
```suggestion
const SEED_RESPONSE_TIMEOUT: u64 = 120_000;
```

PR #150, opened 2022-07-08
Crawler and summary tweaking

zeapoz:
Improves how crawler chooses which nodes to attempt to connect to, also moves summary out of main loop into a separate thread.


src/tools/crawler/main.rs
```diff
@@ -101,6 +106,14 @@ async fn main() {
                 .known_network
                 .nodes()
                 .into_iter()
+                .filter(|(_, node)| {
+                    if let Some(i) = node.last_connected {
+                        if i.elapsed().as_secs() < RECONNECT_INTERVAL {
+                            return false;
+                        }
```
ljedrz, line 113:
```suggestion
                        i.elapsed().as_secs() >= RECONNECT_INTERVAL
```

src/tools/crawler/main.rs
```diff
@@ -101,6 +106,14 @@ async fn main() {
...
+                    }
+                    true
```
ljedrz, line 115:
```suggestion
                    if let Some(i) = node.last_connected {
                        i.elapsed().as_secs() >= RECONNECT_INTERVAL
                    } else {
                        true
                    }
```

src/tools/crawler/network.rs
```diff
@@ -115,6 +119,21 @@ impl KnownNetwork {
...
+        for conn in self.connections() {
+            if conn.last_seen.elapsed().as_secs() > LAST_SEEN_CUTOFF {
+                old_conns.insert(conn);
+            }
+        }
+
+        let mut conns = self.connections.write();
+        for conn in old_conns {
+            conns.remove(&conn);
+        }
```
ljedrz, line 134:
```suggestion
        if !old_conns.is_empty() {
            let mut conns = self.connections.write();
            for conn in old_conns {
                conns.remove(&conn);
            }
        }
```

PR #151, opened 2022-07-11
Make sure summary creation is consistent to loop interval

zeapoz:
Introduces new summary creation loop interval constant and makes sure that the loop happens consistently, regardless of time taken to calculate metrics.

### DIFF
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -1,4 +1,7 @@
-use std::{thread, time::Duration};
+use std::{
+    thread,
+    time::{Duration, Instant},
+};
 
 use clap::Parser;
 use pea2pea::{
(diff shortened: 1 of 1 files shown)


src/tools/crawler/main.rs
```diff
@@ -147,7 +153,13 @@ async fn main() {
...
             }
 
-            thread::sleep(Duration::from_secs(30));
+            if start_time.elapsed() >= Duration::from_secs(SUMMARY_LOOP_INTERVAL) {
+                warn!(parent: crawler.node().span(), "summary calculation took more time than the loop interval");
+            }
+
+            thread::sleep(
+                Duration::from_secs(SUMMARY_LOOP_INTERVAL).saturating_sub(start_time.elapsed()),
+            );
```
ljedrz, line 162: we should bind the result of the subtraction to some new variable instead of using `elapsed` twice here

ljedrz (approved):
LGTM :+1:

PR #155, opened 2022-07-25
Don't panic on non-version messages in perf-tests

zeapoz:
Instead of panicking when node refuses a request from `SyntheticNode`, return an error and abort the handshake. This PR also shuts down nodes after use in `ZG-PERFORMANCE-001`, so we don't clog up the max peer count.


niklaslong (approved):
LGTM 👍

PR #156, opened 2022-07-26
Introduce an RPC server for the crawler

zeapoz:
Implements a basic, mutually exclusive RPC server as an alternative to file logging. The server has one method, `getmetrics`, that responds with a JSON-serialized `NetworkSummary`. Also adds a two cli arguments for binding and enabling the RPC server.

### DIFF
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -1,9 +1,12 @@
 use std::{
+    net::SocketAddr,
+    sync::Arc,
     thread,
     time::{Duration, Instant},
 };
 
 use clap::Parser;
+use parking_lot::Mutex;
 use pea2pea::{
     protocols::{Handshake, Reading, Writing},
     Pea2Pea,
@@ -15,14 +18,16 @@ use tracing_subscriber::filter::{EnvFilter, LevelFilter};
 use ziggurat::{protocol::message::Message, wait_until};
 
 use crate::{
-    metrics::NetworkMetrics,
+    metrics::{NetworkMetrics, NetworkSummary},
     network::KnownNode,
     protocol::{Crawler, MAIN_LOOP_INTERVAL, NUM_CONN_ATTEMPTS_PERIODIC, RECONNECT_INTERVAL},
+    rpc::{initialize_rpc_server, RpcContext},
 };
 
 mod metrics;
 mod network;
 mod protocol;
+mod rpc;
 
 const SEED_WAIT_LOOP_INTERVAL_MS: u64 = 500;
 const SEED_RESPONSE_TIMEOUT_MS: u64 = 120_000;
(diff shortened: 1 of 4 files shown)


src/tools/crawler/main.rs
```diff
@@ -31,10 +36,21 @@ const SUMMARY_LOOP_INTERVAL: u64 = 60;
...
     #[clap(short, long, value_parser, default_value_t = MAIN_LOOP_INTERVAL)]
     crawl_interval: u64,
+
+    /// If present, start an RPC server
+    #[clap(short, long, value_parser)]
+    rpc: bool,
+
+    /// The address to bind the RPC server to
+    #[clap(short, long, value_parser)]
+    bind_rpc: Option<String>,
```
ljedrz, line 53: why not just have a single `rpc_addr` param that's `Option`al and determines if the RPC server should be started? also, can't the address be a `SocketAddr`, i.e. parsed at `clap` level?

src/tools/crawler/main.rs
```diff
@@ -66,6 +82,20 @@ async fn main() {
...
+    let summary_snapshot = Arc::new(Mutex::new(NetworkSummary::default()));
+
+    if args.rpc {
+        // Initialize the RPC server if address is specified.
+        if let Some(addr) = args.bind_rpc {
+            let rpc_addr: SocketAddr = addr.parse().unwrap();
+            let rpc_context = RpcContext::new(Arc::clone(&summary_snapshot));
+            tokio::spawn(async move {
+                initialize_rpc_server(rpc_addr, rpc_context).await;
+            });
```
ljedrz, line 94: any RPC server task should be spawned within `initialize_rpc_server` (and not here)

src/tools/crawler/main.rs
```diff
@@ -145,12 +175,19 @@ async fn main() {
...
+                if !args.rpc {
+                    info!("{}", new_summary);
+                    if let Err(e) = new_summary.log_to_file() {
+                        error!(parent: crawler.node().span(), "Couldn't write summary to file: {}", e);
+                    }
                 }
+
+                // Aquire lock and replace old summary snapshot with the newly generated one.
+                let mut summary_snapshot = summary_snapshot.lock();
+                *summary_snapshot = new_summary;
```
ljedrz, line 190: this should be doable in a single-liner

src/tools/crawler/rpc.rs
```diff
@@ -0,0 +1,49 @@
...
+    let server = HttpServerBuilder::default().build(rpc_addr).await.unwrap();
+
+    let module = create_rpc_module(rpc_context);
+
+    debug!("Starting RPC server at {:?}", server.local_addr().unwrap());
+    let _server_handle = server.start(module).unwrap();
+
+    debug!("RPC server was successfully started");
+
+    std::future::pending::<()>().await;
```
ljedrz, line 36: do we need this? I'd guess it would suffice for the RPC server handle to not go out of scope, which can be achieved if it's kept within the context of `main`

src/tools/crawler/main.rs
```diff
@@ -66,14 +78,24 @@ async fn main() {
...
+    };
 
     crawler.enable_handshake().await;
     crawler.enable_reading().await;
     crawler.enable_writing().await;
 
     for addr in &args.seed_addrs {
         let crawler_clone = crawler.clone();
-        let addr = addr.parse().unwrap();
+        let addr = *addr;
```
ljedrz, line 98: nit: let's just dereference directly at line 109

ljedrz: seems that this avoids the compiler complaining later on; let's leave it be then :+1:

ljedrz (approved):
LGTM :+1:

PR #159, opened 2022-07-27
Add Ctrl-C handling in crawler

zeapoz:
Implements `tokio::signal::ctrl_c` as a future so we can print out `NetworkSummary` once, only on exit, instead of each iteration.


ljedrz (commented):
LGTM with the fmt fix squashed :+1:

ljedrz (approved):

PR #160, opened 2022-07-28
doc: update crawler `README.md`

zeapoz:
Update crawler `README.md` with more information on how it works.

### DIFF
--- a/src/tools/crawler/README.md
+++ b/src/tools/crawler/README.md
@@ -30,7 +30,7 @@ OPTIONS:
 
 ## Metrics
 
-The crawler collects some data for each node it visits and computes some metrics from that data. By default, it will only print and log these on exit (`Ctrl-C`) to a file called `crawler-log.txt`, unless the `--rpc-addr` argument is supplied, in which case these metrics will also be made available to RPC requests.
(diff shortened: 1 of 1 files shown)


src/tools/crawler/README.md
```diff
@@ -22,12 +26,23 @@ OPTIONS:
...
 ```
 
-A sample of the data we collect and metrics we compute (obtained via RPC):
+`--seed-addrs` is the only required argument and needs at least one specified address for it to run.
 
-```
-curl --data-binary '{"jsonrpc": "2.0", "id":0, "method": "ge
+## Metrics
+
+The crawler collects some data for each node it visits and computes some metrics from that data. By default, it will only print and log these on exit (`Ctrl-C`) to a file called `crawler-log.txt`, unless the `--rpc-addr` argument is supplied, in which case these metrics will also be made available to RPC requests.
```
ljedrz, line 33:
```suggestion
The crawler collects some data for each node it visits and compiles metrics based on it. By default, it will only print and log these on exit (`Ctrl-C`) to a file called `crawler-log.txt`, unless the `--rpc-addr` argument is supplied, in which case these metrics will also be made available to RPC requests.
```

ljedrz (approved):
LGTM :+1:

PR #161, opened 2022-08-01
fix: constify header's lengths

lwojtow:


src/protocol/message/constants.rs
```diff
@@ -11,33 +13,35 @@ pub const MAX_MESSAGE_LEN: usize = 2 * 1024 * 1024;
...
-pub const MAGIC_TESTNET: [u8; 4] = [0xfa, 0x1a, 0xf9, 0xbf];
-pub const MAGIC_MAINNET: [u8; 4] = [0x24, 0xe9, 0x27, 0x64];
+pub const MAGIC_TESTNET: [u8; MAGIC_LEN] = [0xfa, 0x1a, 0xf9, 0xbf];
+pub const MAGIC_MAINNET: [u8; MAGIC_LEN] = [0x24, 0xe9, 0x27, 0x64];
 
 #[cfg(test)]
-pub const MAGIC: [u8; 4] = MAGIC_TESTNET;
+pub const MAGIC: [u8; MAGIC_LEN] = MAGIC_TESTNET;
 #[cfg(all(not(test), not(feature = "crawler")))]
 pub const MAGIC: [u8; 4] = MAGIC_MAINNET;
```
ljedrz, line 22: this should use `MAGIC_LEN` too, no?

src/protocol/message/constants.rs
```diff
@@ -11,33 +13,35 @@ pub const MAX_MESSAGE_LEN: usize = 2 * 1024 * 1024;
...
 #[cfg(feature = "crawler")]
-pub const MAGIC: [u8; 4] = MAGIC_MAINNET;
+pub const MAGIC: [u8; MAGIC_LEN] = MAGIC_MAINNET;
+
+pub const COMM_LEN: usize = 12;
```
ljedrz, line 26: probably better to use the full name, so it's more clear wherever it's user

ljedrz (commented):
LGTM with the nits and a `cargo fmt` run

src/protocol/message/constants.rs
lwojtow: Done

src/protocol/message/constants.rs
lwojtow: Done

lwojtow:
I've changed commit message. Should I change branch name as well?

src/protocol/message/mod.rs
```diff
@@ -16,13 +16,15 @@ use crate::protocol::{
     },
 };
 
+pub const MAGIC_LEN: usize = 4;
```
ljedrz, line 19: apologies for not having noticed it before - I think this should go to the `constants.rs` file

ljedrz:
> I've changed commit message. Should I change branch name as well?

Nah, no need; we only care about commit history.

Left one final nit that I missed before, then it's good to go :+1:.

src/protocol/message/mod.rs
lwojtow: Done

ljedrz (approved):
LGTM :+1:

PR #162, opened 2022-08-13
docs: add automake note to README

aphelionz:
This PR adds a note to install `automake` when building `zcashd` from source


niklaslong:
It's listed in the official instructions for `zcashd` that we link to, is there an extra need to specify or can we make that link clearer somehow?

PR #164, opened 2022-09-06
fix: make sure `TestMetrics` doesn't remove state

zeapoz:
This PR fixes a number of issues with our current `DebuggingRecorder` implementation. We now take a snapshot separately from querying the values. This fixes an issue in `resistance::stress_test` where we had two sequential `construct_histogram` queries; the initial one would erase state after its use, leading to nil data in our subsequent query. Now that state is handled separately, we also get a minor boost to performance.

### DIFF
--- a/src/tools/metrics/recorder.rs
+++ b/src/tools/metrics/recorder.rs
@@ -57,19 +57,21 @@ impl Snapshot {
     }
 }
 
-pub fn initialize() -> Snapshotter {
-    let recorder = DebuggingRecorder::new();
-    let snapshotter = recorder.snapshotter();
-    let _ = recorder.install();
-
-    snapshotter
-}
-
 pub struct TestMetrics(Snapshotter);
+
+impl TestMetrics {
+    fn new() -> TestMetrics {
+        let recorder = DebuggingRecorder::new();
+        let snapshotter = recorder.snapshotter();
+        let _ = recorder.install();
+
+        TestMetrics(snapshotter)
+    }
+}
 
 impl Default for TestMetrics {
     fn default() -> Self {
-        Self(initialize())
+        Self::new()
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/tools/metrics/recorder.rs
```diff
@@ -24,44 +26,68 @@ impl Default for TestMetrics {
...
-    pub fn get_val_for(&self, kind: MetricKind, metric: &'static str) -> MetricVal {
-        let key = CompositeKey::new(kind, Key::from_name(metric));
-
-        match &self.0.snapshot().into_hashmap().get(&key).unwrap().2 {
-            DebugValue::Counter(val) => MetricVal::Counter(*val),
-            DebugValue::Gauge(val) => MetricVal::Gauge(val.into_inner()),
-            DebugValue::Histogram(vals) => {
-                MetricVal::Histogram(vals.iter().map(|val| val.into_inner()).collect())
-            }
+    pub fn take_snapshot(&self) -> HashMap<CompositeKey, MetricVal> {
```
niklaslong, line 29: Perhaps we should make this return a new type that then implements `get_counter`, `get_gauge`, etc...?

src/tools/metrics/recorder.rs
```diff
@@ -75,6 +57,42 @@ impl TestMetrics {
     }
 }
 
+pub fn initialize() -> Snapshotter {
```
niklaslong, line 60: Is this used anywhere else but in the `Default` impl for `TestMetrics`?

zeapoz: Not that I know of.

niklaslong: I'd suggest making a `new` function on `TestMetrics` that contains this logic and then have the default impl call `new`.

niklaslong (approved):

PR #167, opened 2022-09-23
Adjust flaky tests to be more consistent

zeapoz:
This PR fixes a few issues with some of the tests that made them unreliable. Notably, `ignore_message_inplace_of_verack` and `bad_checksum` had a few hickups which now should be patched out.

### DIFF
--- a/src/tests/conformance/handshake/ignore_message_inplace_of_verack.rs
+++ b/src/tests/conformance/handshake/ignore_message_inplace_of_verack.rs
@@ -1,6 +1,6 @@
 //! Contains test cases which cover ZG-CONFORMANCE-005 and ZG-CONFORMANCE-006.
 //!
-//! The node ignores non-`Verack` message as a response to initial `Verack` it sent.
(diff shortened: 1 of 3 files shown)


src/tests/conformance/handshake/ignore_message_inplace_of_verack.rs
```diff
@@ -275,7 +263,7 @@ mod when_node_initiates_connection {
         run_test_case(Message::NotFound(block_inv)).await.unwrap();
     }
 
-    /// Checks that `message` gets ignored when sent instead of [`Message::Verack`] when the node
+    /// Checks that `message` doesn't terminate the connection when sent instead of [`Message::Verack`], when the node
     /// initiates the connection.
```
niklaslong, line 267:
```suggestion
    /// Checks that `message` doesn't terminate the connection when sent instead of [`Message::Verack`], when the node
    /// initiates the connection and instead ignores the message.
```

niklaslong (approved):

PR #169, opened 2022-10-06
Improve zcash ci

zeapoz:
This PR improves the GitHub Actions test spinner. It now is able to spin up a node instance and run it through the ziggurat suite, logging results under `results/$TIMESTAMP.jsonl`. Also introduces a separate workflow for running with a stable/nightly compiler.

### DIFF
--- a/.github/workflows/zcashd-nightly.yml
+++ b/.github/workflows/zcashd-nightly.yml
@@ -4,8 +4,6 @@ on:
   workflow_dispatch:    
   push:
     branches: [ "main" ]
-  # pull_request:
-  #   branches: [ main ]
 
 jobs:
   install-zcashd:
--- a/.github/workflows/zcashd-nightly.yml
+++ b/.github/workflows/zcashd-nightly.yml
@@ -3,7 +3,7 @@ name: zcashd-nightly
 on:
   workflow_dispatch:    
   push:
-    branches: [ "zcash-ci" ]
+    branches: [ "main" ]
   # pull_request:
   #   branches: [ main ]
 
--- a/.github/workflows/zcashd-nightly.yml
+++ b/.github/workflows/zcashd-nightly.yml
@@ -1,4 +1,4 @@
-name: zcashd
+name: zcashd-nightly
 
 on:
   workflow_dispatch:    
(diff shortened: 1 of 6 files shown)


aphelionz (approved):
LGTM

.github/workflows/zcashd.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: nightly
```
ljedrz, line 36: we shouldn't target `nightly`, as it's routinely broken

ljedrz: unless we _really_ need some unstable feature?

ljedrz (commented):
I left a question regarding the toolchain use; other than that, I'd squash some of the commits, as the history is messy right now

.github/workflows/zcashd.yml
zeapoz: Well, `--format json` is currently locked behind the nightly compiler. So, unless we want to complicate things further, there's not a whole lot we can do.

muzam1l:
Is this supposed to produce crawler data as well or that's gonna be another pr?

zeapoz:
Yeah, I'd say that is better suited as a follow-up PR

ljedrz (approved):
LGTM :+1:

.github/workflows/zcashd-nightly.yml
```diff
@@ -0,0 +1,97 @@
+name: zcashd-nightly
+
+on:
+  workflow_dispatch:    
+  push:
+    branches: [ "zcash-ci" ]
```
aphelionz, line 6:
```suggestion
    branches: [ "main" ]
```

I think `zcash-ci` won't exist after this PR is merged

aphelionz (approved):
LGTM pending the one comment

.github/workflows/zcashd-nightly.yml
ljedrz: I assumed we'd have a separate branch dedicated to collecting these results, but if not, then 👆

PR #170, opened 2022-10-12
Introduce crawler workflow

zeapoz:
This PR introduces a new CI workflow wherein we crawl the network for ~30 min, logging each run under a new subdirectory: `results/crawler`.

### DIFF
--- a/.github/workflows/crawler.yml
+++ b/.github/workflows/crawler.yml
@@ -59,4 +59,6 @@ jobs:
           # Send SIGINT after 30 min.
           sleep 1800s
           kill -2 $(pidof crawler)
+          # Wait for shutdown to complete.
+          sleep 30s
           cat crawler-log.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/crawler.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: clone zcashd
+        run: git clone https://github.com/zcash/zcash
+      - name: build zcashd
+        run: |
+          cd zcash
+          ./zcutil/build.sh -j$(nproc)
+      - uses: actions/upload-artifact@v2
+        with:
+          name: zcashd-fetch-params
+          path: ./zcash/zcutil/fetch-params.sh
```
aphelionz, line 21: Do we do this twice on purpose? the installation steps?

zeapoz: I couldn't get the node up unless we fetch the params first. So, to avoid uploading the whole `zcash` directory, I just upload the two things we need.

aphelionz (approved):

aphelionz:
Let's give it a spin and see what happens

PR #171, opened 2022-10-17
Introduce ci for zebra

zeapoz:
Similarly to `zcashd`, this PR introduces a new workflow for running `zebra` nodes through the suite. Results are logged as before, but now under a new subdirectory path per node, i.e `results/zcashd` for `zcashd` and `results/zebra` for `zebra`. The PR also migrates the previous trigger-on-push behaviour to schedule runs each day. As an added bonus, this PR also fixes a small issue that prevented the workflow from also logging the `latest.jsonl`.

### DIFF
--- a/.github/workflows/zcashd-nightly.yml
+++ b/.github/workflows/zcashd-nightly.yml
@@ -84,10 +84,10 @@ jobs:
           chmod +x zcash/zcashd
           mkdir -p results/zcashd
           ./ziggurat_test --test-threads=1 --nocapture -Z unstable-options --report-time --format json > results/zcashd/$FILENAME.jsonl
-          rm -f results/zcashd/latest.jsonl
-          cp results/zcashd/$FILENAME.jsonl results/zcashd/latest.jsonl
       - name: git
         run: |
+          rm -f results/zcashd/latest.jsonl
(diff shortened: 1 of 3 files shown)


.github/workflows/zcashd-nightly.yml
```diff
@@ -75,21 +75,23 @@ jobs:
...
           chmod +x zcash/zcashd
-          mkdir -p results
-          ./ziggurat_test --test-threads=1 --nocapture -Z unstable-options --report-time --format json > results/$FILENAME.jsonl
-          rm -r results/latest.jsonl
-          cp results/$FILENAME.jsonl results/latest.jsonl
+          mkdir -p results/zcashd
+          ./ziggurat_test --test-threads=1 --nocapture -Z unstable-options --report-time --format json > result.jsonl
       - name: git
+        env:
+          FILENAME: ${{ github.event.repository.updated_at }}
```
aphelionz, line 87: will `updated_at` return the time of the current run, or the time at which the repo was last updated aka "latest - 1"?

zeapoz: Good question. I'd assume it's the time of the most recent commit, in which case we'll need to find another solution when migrating from trigger-on-push to a cron-based one.

aphelionz (approved):

PR #179, opened 2022-10-28
doc: zebra start command config

danielvladco:


README.md
```diff
@@ -79,11 +79,7 @@ and for Zebra:
 ```toml
 kind = "zebra"
 path = "path/to/zebra/repo"
-start_command = "cargo +stable r -- --verbose start"
-# cargo +stable r   run Zebra using stable Rust
-# --                all args after this will get passed to Zebra
-# verbose           enables verbose logging
-# start             starts the node
+start_command = "target/release/zebrad --verbose"
```
niklaslong, line 82: What's the conclusion on `--verbose`, does it make a difference?

niklaslong (approved):
Other than the pending comment, this looks good to me.

README.md
Rqnsom: And whether we need to use the `start` option or not.
It's funny behaviour if, with no options provided, zebrad defaults to starting the node (which can be achieved with the `start` option.

danielvladco: I don't see any additional logs with `--verbose` when running the tests but maybe it's useful in other scenarios idk. Anyway I can remove it.

niklaslong: I'd suggest keeping `start` and removing `--verbose` in that case. Just to be explicit.

PR #182, opened 2023-01-12
Extend metrics

kylegranger:
This PR extends the crawler metrics that we export.  It also removes some graph-related metrics that are better left to a data crunching app.  The purpose is to expose more data for p2p network visualization.  
This is all a work-in-progress, so many iterations to come.
This is some background info, to explain a bit where this data fits into the Big Picture:
1.  We export a simple adjacency graph, in the AGraph.  This is just of list of nodes connected to a given node.  A 48-bit hash id is created for each node based on the SocketAddr.  This will be used for rough positioning in the renderer, so guaranteed uniqueness is not a requirement.  Betweenness and closeness metrics are also exported.
2. Xcash-gui will still work as should.  Nothing relevant to that has been removed.
3. The basic architecture is **crawler -> data cruncher -> browser renderer**
4.  A prototype data cruncher (rust) is here:  https://github.com/kylegranger/crunchy
It's quite simple:  loads crawler sample, does some magick, spits out state.json
5.  The renderer (TypeScript/nodejs) is here:  https://github.com/kylegranger/showme
Right now, it loads and parses the state.json file created by crunchy.  Not yet displaying anything yet (other than textured background image).
6.  Regarding the crawler:  there are still lots of changes we can do there.  This current PR still does the response via JSON-RPC.  Future Stuff I've already talked about with Niklas:
- writing to file rather than response, adding filepath to request params
- different flags, features (e.g., optionally write the SocketAddrs)
- not create an agraph at all, and writing out the edges -- much discussion about that
- writing out the adjacency matrix (DMatrix)
7.  One note regarding `spectre`:  I link to the latest repo version, and not what is currently published in `crates.io`.  Much of the work for getting this all up and running happened in that library; namely, the betweenness and closeness centrality computations.


src/protocol/payload/mod.rs
```diff
@@ -133,7 +133,7 @@ impl Codec for VarInt {
             len @ 0x00..=0xfc => len as u64,
             0xfd => u16::from_le_bytes(read_n_bytes(bytes)?) as u64,
             0xfe => u32::from_le_bytes(read_n_bytes(bytes)?) as u64,
-            0xff => u64::from_le_bytes(read_n_bytes(bytes)?) as u64,
+            0xff => u64::from_le_bytes(read_n_bytes(bytes)?),
```
kylegranger, line 136: not my code, but clippy complained about superfluous case

asmie (approved):
LGTM.
Instead of just merging this PR use squash first as there are 19 commits here :)

Rqnsom (commented):
I can't provide a good review; I'll have to study the crawler code first.
But:
`crawler -> data cruncher -> browser renderer`
Will cruncher/renderer end up under the ziggurat repo eventually?

niklaslong (approved):

Cargo.toml
```diff
@@ -12,14 +12,15 @@ hex = "0.4.3"
...
+md5 = "0.7.0"
 metrics = "0.20.0"
 metrics-util = "0.14.0"
 parking_lot = "0.12"
 pea2pea = "0.40"
 rand = "0.8"
 rand_chacha = "0.3"
 sha2 = "0.10"
-spectre = "0.4.0"
+spectre = { git = "https://github.com/niklaslong/spectre" }
```
niklaslong, line 23: Perhaps depend on a specific commit here (i.e. using `rev`) so it doesn't break immediately if we make breaking changes to spectre?

PR #187, opened 2023-01-24
refactor: use IpAddr only to represent a connection

lwojtow:
* Use newest ziggurat-core-crawler
* Refactor out method to add new nodes based on connections


asmie (approved):
LGTM

PR #188, opened 2023-01-27
chore: update dependencies

lwojtow:


asmie (approved):
LGTM

Rqnsom (commented):
Shouldn't we add support for a `start_listening()` from pea2pea v0.43 method for the SyntheticNode?
For those tests in which the node connects to our synth node?
Are all tests passing with this commit? Is C002 passing? Could we double-check?

otherwise  LGTM

lwojtow:
> Shouldn't we add support for a `start_listening()`
it turned out  to be a lot more convoluted that that. It is expected that listening_addr is present in many places, so the easiest fix was to add start_listening when building the node.

PR #189, opened 2023-02-17
deps(ci): remove `actions-rs`

zeapoz:
Replaces all the `actions-rs` actions with standard `rustup/cargo` commands now that `actions-rs` is no longer maintained.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - run: rustup toolchain install stable --profile minimal
+      - run: rustup toolchain install nightly --profile minimal
       - uses: Swatinem/rust-cache@v2
       - run: cargo fmt --all -- --check
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,78 +11,38 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
+      - run: rustup toolchain install stable --profile minimal
       - uses: Swatinem/rust-cache@v2
-      - uses: actions-rs/cargo@v1
-        with:
-          command: check
-          args: --all-targets
+      - run: cargo check --all-targets
 
   test:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
+      - run: rustup toolchain install stable --profile minimal
       - uses: Swatinem/rust-cache@v2
-      - uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: -- --test-threads=1 --ignored
-
+      - run: cargo test -- --test-threads=1 --ignored
 
   clippy:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-          components: clippy
+      - run: rustup toolchain install stable --profile minimal
       - uses: Swatinem/rust-cache@v2
-      - uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: --all-targets -- -D warnings
-
+      - run: cargo clippy --all-targets -- -D warnings
 
   rustfmt:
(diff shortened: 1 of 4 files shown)


lwojtow (approved):

.github/workflows/test.yml
```diff
@@ -11,78 +11,38 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly
-          override: true
-          components: rustfmt
+      - run: rustup toolchain install stable --profile minimal
```
niklaslong, line 38: I don't think the minimal profile includes `rustfmt`? https://rust-lang.github.io/rustup/concepts/profiles.html

.github/workflows/test.yml
```diff
@@ -11,78 +11,38 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-          components: clippy
+      - run: rustup toolchain install stable --profile minimal
```
niklaslong, line 30: ditto for clippy

niklaslong:
Re-minimal toolchain with the missing components: I wonder if it works because the toolchain is being cached from previous runs?

zeapoz:
> Re-minimal toolchain with the missing components: I wonder if it works because the toolchain is being cached from previous runs?

This is from some testing I did:

```
Run rustup toolchain install stable --profile minimal
info: syncing channel updates for 'stable-x86_64-unknown-linux-gnu'
info: latest update on 2023-02-09, rust version 1.67.1 (d5a82bbd2 2023-02-07)
info: downloading component 'clippy'
info: downloading component 'rustfmt'
info: downloading component 'cargo'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: removing previous version of component 'clippy'
info: removing previous version of component 'rustfmt'
info: removing previous version of component 'cargo'
info: removing previous version of component 'rust-std'
info: removing previous version of component 'rustc'
info: installing component 'clippy'
info: installing component 'rustfmt'
info: installing component 'cargo'
info: installing component 'rust-std'
info: installing component 'rustc'
```

It's definitely cached, but the minimal profile also seems to download `rustfmt` and `clippy`, not quite sure what to make of that.

niklaslong:
Yeah, only took a quick look at the `rustup` repo but couldn't find anything conclusive. This is probably fine for now but if it starts breaking we'll know why.

niklaslong:
On second thought, let's maybe specify the component flag: https://rust-lang.github.io/rustup/concepts/components.html

.github/workflows/test.yml
```diff
@@ -11,78 +11,38 @@ jobs:
...
-          override: true
-          components: rustfmt
+      - run: rustup toolchain install stable --profile minimal
       - uses: Swatinem/rust-cache@v2
-      - uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
-
+      - run: cargo fmt --all -- --check
```
Rqnsom, line 40: ...can I ask if this runs with the nightly option? I don't know how to read `.yml` files, but I see line 65 (old file) specified nightly toolchain

niklaslong: good catch! nope, this'll run on the `stable` toolchain with the changes.

niklaslong: It's the same in a few other sections of the workflow (speaking of, do we still need the `nightly` test run?).

zeapoz: Yeah, so I replaced some of the toolchains with the stable variant where no nightly options were used (from what I could tell at least).

zeapoz: Was there a particular reason for using nightly on `cargo fmt`?

niklaslong: We have fmt config that uses nightly options: https://github.com/runziggurat/zcash/blob/main/.rustfmt.toml
That being said, I don't know if they're still only available on nightly, perhaps they've been stabilised in the interim, worth checking.

zeapoz: Does not look to be part of stable just yet. I'll push up a fix, it's only `cargo fmt` that needs nightly, right?

niklaslong: should be yes, stable `clippy` is fine 👍 
We had some `nightly` workflows (tests and all) there too, not sure if they are still needed.

zeapoz: I think the only actual necessary ones are for `cargo fmt` & for accessing the `-Z` flags during testing, so we should be all good. Ready for merge you think?

niklaslong (approved):

PR #190, opened 2023-02-22
feat(ci): compress results

zeapoz:
This PR modifies the test runners + crawler workflow to compress the results before committing. This **should not** affect the GUI, given that the latest result (`latest.jsonl`) will still be kept uncompressed.


Rqnsom (approved):
LGTM! looks nice

could we also gzip all the existing non-compressed files in a separate PR?
considering the current repo size is huge

zeapoz:
Coming right up :rocket:

PR #192, opened 2023-02-23
feat: update crawler to use latest NetworkSummary & spectre

kylegranger:
I've updated the zcash crawler to use:
- new spectre API to fetch indices
- remove AGraph as struct, using updated NetworkSummary

I have run the zcash crawler per normal, and got a correctly formed sample.json back (e.g., using the `indices` field in NetworkSummary, rather than `agraph`.


zeapoz (approved):
LGTM :+1:

PR #195, opened 2023-03-06
feat: increase crawler's max response size from 20MB to 200MB

kylegranger:
This should solve the problem discovered by Piotr with empty crawler responses due to response size exceeding server maximum.

### DIFF
--- a/src/tools/crawler/rpc.rs
+++ b/src/tools/crawler/rpc.rs
@@ -7,8 +7,8 @@ use ziggurat_core_crawler::summary::NetworkSummary;
 
 pub struct RpcContext(Arc<Mutex<NetworkSummary>>);
 
-/// Make this now 200MB
-pub const MAX_RESPONSE_SIZE: u32 = 200000000;
+/// Allow JSON-RPC response size to be up to 200MB
+pub const MAX_RESPONSE_SIZE: u32 = 200_000_000;
 
 impl RpcContext {
     /// Creates a new RpcContext.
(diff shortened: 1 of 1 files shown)


src/tools/crawler/rpc.rs
```diff
@@ -7,6 +7,9 @@ use ziggurat_core_crawler::summary::NetworkSummary;
 
 pub struct RpcContext(Arc<Mutex<NetworkSummary>>);
 
+/// Make this now 200MB
```
zeapoz, line 10:
```suggestion
/// Allow JSON-RPC response size to be up to 200MB.
```

src/tools/crawler/rpc.rs
```diff
@@ -7,6 +7,9 @@ use ziggurat_core_crawler::summary::NetworkSummary;
...
+pub const MAX_RESPONSE_SIZE: u32 = 200000000;
```
zeapoz, line 11:
```suggestion
pub const MAX_RESPONSE_SIZE: u32 = 200_000_000;
```

src/tools/crawler/rpc.rs
kylegranger: 👍  done

Rqnsom (commented):
LGTM

lwojtow (approved):

src/tools/crawler/rpc.rs
kylegranger: done

zeapoz (approved):
LGTM :+1:

PR #196, opened 2023-03-06
fix: always procure crawler runtime last

zeapoz:
`crawler_runtime` is currently calculated before a function (`get_filtered_adjacency_indices`) that takes a long time to compute, resulting in skewed runtime results. This PR makes sure that `crawler_runtime` only gets procured when returning `NetworkSummary`, and adds a small `info!` line in the summary calculation loop for displaying time taken to run the calculations.


Rqnsom (approved):
nice!

PR #197, opened 2023-03-07
Fail crawler workflow if an error code is returned

zeapoz:
- Grep for `error` in `latest.json` and fail workflow if it is found.
- Move some commands into a new step for processing data for added clarity.


kylegranger (approved):
LGTM

PR #199, opened 2023-03-09
feat: address serving based on peer list

asmie:
This commit adds new feature which is serving addresses upon requested based on the Intelligent Peer Sharing results.

Peer list need to be taken from command line and if there is no list provided no addresses will be served to the nodes.

If peer list is provieded and is valid, crawler will send proposed peers when will receive getaddr command from node it is crawling at the moment. If there are any peers to be proposed they are send as addr message (limited to 1000).

Peer list would not be proposed if node will not ask about them. Crawler won't check also if proposed peers has been accepted or ignored.


src/tools/crawler/main.rs
```diff
@@ -77,7 +84,14 @@ async fn main() {
...
     // Create the crawler with the given listener address.
-    let crawler = Crawler::new().await;
+    let mut crawler = Crawler::new().await;
+
+    if let Some(peer_file) = args.peer_file {
+        if let Ok(js) = fs::read_to_string(peer_file) {
+            crawler.peer_list = serde_json::from_str(&js).expect("failed to parse peer file");
+            info!(parent: crawler.node().span(), "loaded peer list for {:?} nodes", crawler.peer_list.len());
+        }
+    }
```
Rqnsom, line 94: so the crawler will read the file only once and always suggest the same peer list throughout its runtime? 🕵️‍♀️

src/tools/crawler/main.rs
```diff
@@ -77,7 +84,14 @@ async fn main() {
     let args = Args::parse();
 
     // Create the crawler with the given listener address.
-    let crawler = Crawler::new().await;
+    let mut crawler = Crawler::new().await;
+
+    if let Some(peer_file) = args.peer_file {
+        if let Ok(js) = fs::read_to_string(peer_file) {
+            crawler.peer_list = serde_json::from_str(&js).expect("failed to parse peer file");
+            info!(parent: crawler.node().span(), "loaded peer list for {:?} nodes", crawler.peer_list.len());
```
Rqnsom, line 92: should we ensure the file has a recent algo output

Rqnsom (commented):
I'm not sure about this one yet - this should be tested more with some real networks before we release it

src/tools/crawler/main.rs
asmie: I think so.
The main flow is that crawler works and creates sample file for crunchy. So the crawler need to finish its job. Crunchy is taking that file and (if the network is big) after some hours produces peer proposal to be used as next crawler run. 
And that how it goes...

I can't see right now any way to increase dynamics as the more crawler is running the more nodes it discovers. Discovering nodes changes the network and graphs and therefore, it could change the ways crunchy and IPS are working. 

I think that it would be safe to work over some completed state - so crunchy works over completed network crawl and crawler works having last stable peerlist.

Rqnsom: hmhm, maybe we can change so that the peer lists here... are suggested only once - like consumed!

maybe, we could have an automatic cycle like this:
```
1. run crawler
 <- no input
 -> continues running for N amount of time
 -> outputs the network state and dies

loop {
    2. run crunchy
     <- consumes the latest current state
     -> outputs IPS the algo peer lists
     
    3. run crawler
     <- applies (consumes) the latest peer lists
     -  continues running for N amount of time
     -> outputs the network state and dies
}
```

Rqnsom: > hmhm, maybe we can change so that the peer lists here... are suggested only once - like consumed!

or we just make it so it periodically checks for some fresh IPS suggestion file, once it finds it - it applies and then waits for a new file to be generated by crunchy?

PR #200, opened 2023-03-14
feat(ci): display diff from last run

zeapoz:
Included with this PR (identical to the `XRPL` one), for both `zcashd` and `zebra`:
- A new step that will display the output of the diff command between the workflows latest run (latest.jsonl) and the one prior to that (comparison.jsonl)
- The json files are modified in a new function, filter_json, to filter out things that would clog the output of diff, e.g exec_time and type.


kylegranger (approved):
LGTM

PR #202, opened 2023-03-20
Introduce more CI reusability

zeapoz:
- New `install-zcashd` workflow
- Split `process-results` into, `process-results` and `diff-with-previous`
- Modifies `process-results` to be compatible with calls from crawler workflows

### DIFF
--- a/.github/workflows/install-zcashd.yml
+++ b/.github/workflows/install-zcashd.yml
@@ -1,0 +1,21 @@
+on:
+  workflow_call:
+
+jobs:
+  install-zcashd:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone zcashd
+        run: git clone https://github.com/zcash/zcash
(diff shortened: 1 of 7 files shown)


.github/workflows/install-zcashd.yml
```diff
@@ -0,0 +1,21 @@
+on:
+  workflow_call:
+
+jobs:
+  install-zcashd:
```
asmie, line 5: Unification case: targets for zcash are called `install-zcashd` which in fact does building also and for zebra there is a target `build-zebra` which does the same (so building and uploading artifacts). 
Both should be unified (but maybe that's the plan for zebra) 😄

asmie (approved):
👍

.github/workflows/install-zcashd.yml
zeapoz: Done

PR #203, opened 2023-03-21
Allow different file extensions in `process-results`

zeapoz:
Crawler workflows uses `.json` files, while suite workflows uses `.jsonl` files. This PR allows both extensions to be used in the `process-results` workflow.

### DIFF
--- a/.github/workflows/process-results.yml
+++ b/.github/workflows/process-results.yml
@@ -4,24 +4,30 @@ on:
       name:
         required: true
         type: string
+      extension:
+        type: string
+        default: "jsonl"
 
 jobs:
   process-results:
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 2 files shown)


.github/workflows/process-results.yml
```diff
@@ -4,24 +4,32 @@ on:
...
       NAME: ${{ inputs.name }}
+      EXTENSION: ${{ inputs.extension }}
     steps:
       - uses: actions/checkout@v3
       - uses: actions/download-artifact@v3
         with:
           name: latest-result
-          path: results/$NAME/latest.jsonl
+          path: ./
+      - name: Replace with latest result
```
Rqnsom, line 23:
```suggestion
      - name: Replace with the latest result
```

Rqnsom (approved):
lgtm

PR #210, opened 2023-03-28
feat: node network checker

asmie:
Allow recognize type of network for each node. Currently it allows to check if node is from real ZCash network or from any other.

Classification is based on three factors:
start_height - this MUST match
port
agent string

If either port or agent string would not match node still can be considered as ZCash node as long as start_height matches. The only exception is when agent string identifies node as version 6 (which is other network than ZCash) - that is blocker and even when two other conditions are met, node wouldn't be considered as valid ZCash node.

### DIFF
--- a/src/tools/crawler/metrics.rs
+++ b/src/tools/crawler/metrics.rs
@@ -36,6 +36,13 @@ impl NetworkMetrics {
     }
 }
 
+// Updates the node's network type.
+//
+// The decision of whether a node belongs to the Zcash network is based on these factors:
+//  `start_height` - this MUST match
+// and one of the additional factors also must match as an extra confirmation:
+// `port`
+// `agent`
 fn recognize_network_types(
     nodes: &HashMap<SocketAddr, KnownNode>,
     good_nodes: &Vec<SocketAddr>,
@@ -76,14 +83,15 @@ fn recognize_network_types(
             agent_matches = true;
         }
 
+        // Check if the height is alright - this is a mandatory check for any zcash node implementation.
         let height = nodes[node].start_height.unwrap_or(0);
         if height < MIN_BLOCK_HEIGHT {
             node_network_types.push(NetworkType::Unknown);
             continue;
         }
 
-        // Height must match and either port or agent must match to recognize node as zcash node and
-        // there were no conditions that explicitly blocks matching.
+        // When a block height is correct, we still need one additional confirmation:
+        // In rare cases, the agent or the port won't use a commonly used value.
         if port_matches || agent_matches {
             node_network_types.push(NetworkType::Zcash);
         } else {
--- a/src/tools/crawler/metrics.rs
+++ b/src/tools/crawler/metrics.rs
@@ -1,9 +1,13 @@
 use std::{collections::HashMap, net::SocketAddr};
 
+use regex::Regex;
 use spectre::{edge::Edge, graph::Graph};
-use ziggurat_core_crawler::summary::NetworkSummary;
+use ziggurat_core_crawler::summary::{NetworkSummary, NetworkType};
 
 use crate::{network::LAST_SEEN_CUTOFF, Crawler};
+
+const MIN_BLOCK_HEIGHT: i32 = 2_000_000;
+const ZCASH_P2P_PORT: u16 = 8233;
 
 #[derive(Default)]
 pub struct NetworkMetrics {
@@ -48,7 +52,7 @@ pub fn new_network_summary(crawler: &Crawler, graph: &Graph<SocketAddr>) -> Netw
     let mut protocol_versions = HashMap::with_capacity(num_known_nodes);
     let mut user_agents = HashMap::with_capacity(num_known_nodes);
 
-    for (_, node) in nodes {
+    for (_, node) in nodes.clone() {
         if node.protocol_version.is_some() {
             protocol_versions
                 .entry(node.protocol_version.unwrap().0)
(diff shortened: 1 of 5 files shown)


src/tools/crawler/metrics.rs
```diff
@@ -61,6 +65,56 @@ pub fn new_network_summary(crawler: &Crawler, graph: &Graph<SocketAddr>) -> Netw
...
+            port_matches = true;
+        }
+
+        let agent = if let Some(agent) = &nodes[node].user_agent {
+            agent.0.clone()
+        } else {
+            "".to_string()
+        };
+        let zcash_re = Regex::new(r"^/MagicBean:(\d)\.(\d)\.(\d)/$").unwrap();
+        let zebra_re = Regex::new(r"^/Zebra:(\d)\.(\d)\.(\d)").unwrap();
```
Rqnsom, line 81: this provides more context below
```suggestion
        let zcash_regex = Regex::new(r"^/MagicBean:(\d)\.(\d)\.(\d)/$").unwrap();
        let zebra_regex = Regex::new(r"^/Zebra:(\d)\.(\d)\.(\d)").unwrap();
```

src/tools/crawler/metrics.rs
```diff
@@ -61,6 +65,56 @@ pub fn new_network_summary(crawler: &Crawler, graph: &Graph<SocketAddr>) -> Netw
...
     }
 
+    let mut node_network_types = Vec::with_capacity(num_good_nodes);
+    for node in &good_nodes {
+        let mut port_matches = false;
+        let mut agent_matches = false;
+        let mut height_matches = false;
+        let mut blocker = false;
+
+        if node.port() == ZCASH_P2P_PORT {
```
kylegranger, line 75: don't really need mut?
```
let port_matches = node.port() == ZCASH_P2P_PORT;
```

src/tools/crawler/metrics.rs
```diff
@@ -61,6 +65,56 @@ pub fn new_network_summary(crawler: &Crawler, graph: &Graph<SocketAddr>) -> Netw
...
+        }
+
+        let cap_ze = zebra_re.captures(agent.as_str());
+        if cap_ze.is_some() {
+            // Accept all zebra versions
+            agent_matches = true;
+        }
+
+        let height = nodes[node].start_height.unwrap_or(0);
+        if height > MIN_BLOCK_HEIGHT {
```
kylegranger, line 105: same here
```
let height_matches = height > MIN_BLOCK_HEIGHT;
```

src/tools/crawler/metrics.rs
```diff
@@ -48,7 +52,7 @@ pub fn new_network_summary(crawler: &Crawler, graph: &Graph<SocketAddr>) -> Netw
     let mut protocol_versions = HashMap::with_capacity(num_known_nodes);
     let mut user_agents = HashMap::with_capacity(num_known_nodes);
 
-    for (_, node) in nodes {
+    for (_, node) in nodes.clone() {
```
Rqnsom, line 55: can we use `iter()` here? 🕵️‍♀️

kylegranger (approved):
the criteria are reasonable, and the logic is sound.

src/tools/crawler/metrics.rs
```diff
@@ -61,6 +65,56 @@ pub fn new_network_summary(crawler: &Crawler, graph: &Graph<SocketAddr>) -> Netw
...
+        let zebra_re = Regex::new(r"^/Zebra:(\d)\.(\d)\.(\d)").unwrap();
+
+        let cap_zc = zcash_re.captures(agent.as_str());
+        if let Some(cap) = cap_zc {
+            let major = cap.get(1).unwrap().as_str().parse::<u32>().unwrap();
+            if major < 6 {
+                // Accept all zcash versions < 6 (6 is Flux)
+                agent_matches = true;
+            } else if major == 6 {
+                blocker = true; // Block all zcash versions 6 (Flux) even if they are on the right port
```
Rqnsom, line 90: a general comment regarding this part and the line 105:

if the must-condition is not met (`height is not okay` || `major==6`), we can just call
```rust
node_network_types.push(NetworkType::Unknown);
continue;
```

so the logic on line 111 gets easier to read.
also we will btw remove  those extra variables `height_matches` and `blocker`

src/tools/crawler/metrics.rs
asmie: Done.

src/tools/crawler/metrics.rs
asmie: Done.

src/tools/crawler/network.rs
```diff
@@ -23,6 +23,8 @@ pub struct KnownNode {
     pub protocol_version: Option<ProtocolVersion>,
     /// The node's user agent.
     pub user_agent: Option<VarStr>,
+    /// The node's best block height.
```
Rqnsom, line 26: the best block height? 😃

src/tools/crawler/metrics.rs
```diff
@@ -61,6 +65,56 @@ pub fn new_network_summary(crawler: &Crawler, graph: &Graph<SocketAddr>) -> Netw
         }
     }
 
+    let mut node_network_types = Vec::with_capacity(num_good_nodes);
+    for node in &good_nodes {
+        let mut port_matches = false;
+        let mut agent_matches = false;
```
Rqnsom: alternative idea:
```suggestion
        let mut port_ok = false;
        let mut agent_ok = false;
```

Rqnsom (approved):
lgtm! 👍

src/tools/crawler/metrics.rs
asmie: Changed.
`re` is generally widely used across Regex library (and samples) to indicate if something is regular expression but changed if that wasn't clear enough.

src/tools/crawler/metrics.rs
asmie: Yes, we can. Then `user_agent` needs to be cloned.
Done.

src/tools/crawler/network.rs
asmie: Probably was thinking about something else 😃  Changed.

src/tools/crawler/metrics.rs
```diff
@@ -61,6 +65,50 @@ pub fn new_network_summary(crawler: &Crawler, graph: &Graph<SocketAddr>) -> Netw
...
+            if major < 6 {
+                // Accept all zcash versions < 6 (6 is Flux)
+                agent_matches = true;
+            } else if major == 6 {
+                blocker = true; // Block all zcash versions 6 (Flux) even if they are on the right port
+            }
+        }
+
+        let cap_ze = zebra_re.captures(agent.as_str());
+        if cap_ze.is_some() {
```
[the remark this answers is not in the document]
asmie: Added.

src/tools/crawler/metrics.rs
Rqnsom: >  re is generally widely used across Regex library (and samples) to indicate if something is regular expression but changed if that wasn't clear enough.

aha! I wasn't aware. you could have kept it, I would get used to it if it's commonly used elsewhere

src/tools/crawler/metrics.rs
```diff
@@ -29,6 +36,64 @@ impl NetworkMetrics {
...
+
+        // Look for zebra agent like "/Zebra:1.0.0-rc.4/"
+        let cap_ze = zebra_regex.captures(agent.as_str());
+        if cap_ze.is_some() {
+            // Accept all zebra versions
+            agent_matches = true;
+        }
+
+        let height = nodes[node].start_height.unwrap_or(0);
+        if height < MIN_BLOCK_HEIGHT {
```
Rqnsom, line 80:
```suggestion
 
        // Check if the height is alright - this is a mandatory check for any zcash node implementation.
        let height = nodes[node].start_height.unwrap_or(0);
        if height < MIN_BLOCK_HEIGHT {
```

src/tools/crawler/metrics.rs
```diff
@@ -29,6 +36,64 @@ impl NetworkMetrics {
...
+            node_network_types.push(NetworkType::Unknown);
+            continue;
+        }
+
+        // Height must match and either port or agent must match to recognize node as zcash node and
+        // there were no conditions that explicitly blocks matching.
```
Rqnsom, line 86:
```suggestion
        // When a block height is correct, we still need one additional confirmation:
        // In rare cases, the agent or the port won't use a commonly used value. 
```

Rqnsom (approved):
LGTM, I would just add a few extra comments to clarify

(these are basically copied from the PR description we have now, but once we merge it, it won't be so accessible to future readers)

src/tools/crawler/metrics.rs
Rqnsom: > Probably depends on one's preferences 

that's so true!
that's why I asked for some extra comments which provide more context while keeping things simple 😃

src/tools/crawler/metrics.rs
asmie: Changed.

src/tools/crawler/metrics.rs
asmie: Changed.

Rqnsom (approved):

PR #212, opened 2023-03-30
Introduce Nix integration :snowflake: 

zeapoz:
- Introduces Nix:
  - `flake.nix`-file providing a reproducible CI/CD/Dev environment
  - A shell hook is also provided to setup stable rust as the default, and installing the nightly `rustfmt` component
  - `flake.lock`-file for dependency pinning (will be very useful for caching in the future)
- Updates workflows:
  - `check-and-lint` now uses a matrix job and the provided `flake.nix`-shell

### DIFF
--- a/.github/workflows/check-and-lint.yml
+++ b/.github/workflows/check-and-lint.yml
@@ -8,16 +8,23 @@ on:
 
 jobs:
   core:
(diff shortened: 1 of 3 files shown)


kylegranger (approved):
LGTM  👍

PR #213, opened 2023-04-05
feat: development tests

zeapoz:
This PR introduces development tests for tinkering, similar to the ones already existing in XRPL.

The tests have support for a number of different configuration options, these are:
- enable/disable node's logs to stdout
- enable/disable tracing
- enable/disable attaching a single synthetic node to the node
  - suboption: choose the initiator for the connection


Rqnsom (approved):
the periodic crawler functionality is missing, but can be added later

LGTM 🚀

PR #216, opened 2023-04-12
More info in the logs about the new connection during handshake

Rqnsom:
```
    chore: log connection info during the handshake
    
    We should be able to see more info in the logs regarding our new
    connection.
```

A second and an extra commit which I needed for some local tests:
```
    feat(synth_node): add disconnect function
```


kylegranger (approved):
verbosity is good 👍

asmie (approved):
👍

PR #218, opened 2023-04-18
Add new environment variables to push to gcs

rodriguezcondec:


kylegranger (approved):
👍

aphelionz (approved):

Rqnsom (approved):

zeapoz (changes requested):
Changes are looking good to me, but these lines will also need to be changed (in each workflow).
https://github.com/runziggurat/zcash/blob/a5519364f595b8e637ae45bac239168fea0247e5/.github/workflows/zcashd-nightly.yml#L52-L63

zeapoz:
You can probably just remove the uploading of the previous results and instead directly fetch `latest.$FILENAME` here https://github.com/runziggurat/ziggurat-core/blob/1de5c466c4e0216506aaf2b76b291c684f439aaf/.github/workflows/diff-with-previous.yml#L11-L13

rodriguezcondec:
> Changes are looking good to me, but these lines will also need to be changed (in each workflow).
> 
> https://github.com/runziggurat/zcash/blob/a5519364f595b8e637ae45bac239168fea0247e5/.github/workflows/zcashd-nightly.yml#L52-L63

I can take a look but I don't think this would break now the CI right? It would just create a `results/` folder within the github actions virtual machine instance to run `./ziggurat_test` and then upload these results, 🤔 so in theory, we could leave it as it is and it should work right? @zeapoz

rodriguezcondec:
> then fail since there no

Yeah, you're right, will open a PR to change that. I can directly remove that step and that would be it? Or do i need to include https://github.com/runziggurat/ziggurat-core/blob/1de5c466c4e0216506aaf2b76b291c684f439aaf/.github/workflows/diff-with-previous.yml#L11-L13 as well?

PR #219, opened 2023-04-18
doc: init ci/cd documentation

zeapoz:


.github/workflows/README.md
```diff
@@ -0,0 +1,33 @@
+# Ziggurat CI/CD
```
Rqnsom, line 1: One comment!
All this info is great! But it feels more like a "Daily CI/CD"

Could we also mention and describe somewhere per-commit CI flow? (another PR maybe)

Rqnsom (approved):
LGTM with a small comment

.github/workflows/README.md
zeapoz: Yeah of course, that sort of ties in with some small Nix documentation that I want to provide as well, so I'll include it with that PR :+1:

PR #221, opened 2023-04-25
feat: crawler now accepts DNS seeder argument

asmie:
There is now possibility to add DNS seeders as arguments for crawler:
`-d, --dns-seed <DNS_SEED_ADDRS>`
They can be used to obtain IP addresses for initial nodes. DNS seeder can be used instead of local node for crawling as this feature does basically the same - plain node is also checking DNS seeders upon start.


src/tools/crawler/main.rs
```diff
@@ -75,10 +80,34 @@ fn start_logger(default_level: LevelFilter) {
...
+        seed_addrs.extend_from_slice(&seeds);
+    }
+
+    if let Some(dns_seed) = args.dns_seed {
+        for seed_str in dns_seed {
+            let response = lookup_host(&seed_str);
+
+            if let Ok(response) = response {
+                for address in response.iter() {
+                    seed_addrs.push(SocketAddr::new(*address, 8233)); // DNS addrs use this port
```
kylegranger, line 98: maybe make this number a const

kylegranger (approved):
👍  this should prove to be useful

src/tools/crawler/main.rs
asmie: Right, it's been already defined but wasn't public. 
Changed. 👍

src/tools/crawler/main.rs
```diff
@@ -39,8 +40,12 @@ const LOG_PATH: &str = "crawler-log.txt";
 #[clap(author, version, about, long_about = None)]
 struct Args {
```
zeapoz, line 41: How about using an `ArgGroup` to make sure that at least one of the source addrs arguments is given?
https://docs.rs/clap/latest/clap/struct.ArgGroup.html#method.required

This should do the job:
```suggestion
#[clap(author, version, about, long_about = None)]
#[clap(group(
        ArgGroup::new("source_addrs")
            .required(true)
            .args(&["seed-addrs", "dns_seed"])
        ))],
struct Args {
```

zeapoz (approved):
Looking very good! I can update the crawler workflow once this lands :+1:

src/tools/crawler/main.rs
asmie: Nice one!
Exact form would be like that:

```
#[clap(group(
ArgGroup::new("source_addrs")
.required(true)
.multiple(true)
.args(&["seed_addrs", "dns_seed"])
))]
```

I've added `.multiple(true)` which allows to set both dns seeder and give just peer 😄

src/tools/crawler/README.md
```diff
@@ -12,21 +12,24 @@ $ cargo run --release --features crawler --bin crawler -- --help
...
+            
     -h, --help
             Print help information
 
     -r, --rpc-addr <RPC_ADDR>
             If present, start an RPC server at the specified address
 
     -s, --seed-addrs <SEED_ADDRS>...
-            The initial addresses to connect to
+            The initial addresses to connect to (can be used multiple times)
```
Rqnsom, line 26: To make the cmd UI simpler (with fewer options) - an alternative solution could have been to allow this argument to receive both (but then we would have used the` Vec<String>` and parsed/checked the IP locally):
 - human-friendly domain names (which we would resolve internally with a DNS service)
 - raw alphanumeric IP address

src/tools/crawler/main.rs
```diff
@@ -37,10 +38,20 @@ const LOG_PATH: &str = "crawler-log.txt";
 
 #[derive(Parser)]
 #[clap(author, version, about, long_about = None)]
+#[clap(group(
+ArgGroup::new("source_addrs")
+.required(true)
+.multiple(true)
+.args(&["seed_addrs", "dns_seed"])
```
Rqnsom, line 45: I'm gonna bookmark this part 😃

src/tools/crawler/metrics.rs
```diff
@@ -10,7 +10,7 @@ use crate::{
 };
 
 const MIN_BLOCK_HEIGHT: i32 = 2_000_000;
-const ZCASH_P2P_PORT: u16 = 8233;
+pub const ZCASH_P2P_PORT: u16 = 8233;
```
Rqnsom, line 13: the default part implies (for more context) it's not necessarily always the 8233
```suggestion
pub const ZCASH_P2P_DEFAULT_PORT: u16 = 8233;
```

src/tools/crawler/main.rs
```diff
@@ -75,10 +86,34 @@ fn start_logger(default_level: LevelFilter) {
...
+        for seed_str in dns_seed {
+            let response = lookup_host(&seed_str);
+
+            if let Ok(response) = response {
+                for address in response.iter() {
+                    seed_addrs.push(SocketAddr::new(*address, ZCASH_P2P_PORT)); // DNS addrs use this port
+                    info!(parent: crawler.node().span(), "DNS seed {} address added: {}", seed_str, address);
+                }
+            } else {
+                error!(parent: crawler.node().span(), "DNS seed {} lookup failed: {}", seed_str, response.err().unwrap());
```
Rqnsom, line 108: Maybe we could kill the process here, so we are aware if this ever happens - otherwise, the mistake might go unnoticed by the user (who might have accidentally entered a wrong domain name)

and the user can be our CI e.g.

src/tools/crawler/main.rs
```diff
@@ -75,10 +86,34 @@ fn start_logger(default_level: LevelFilter) {
...
+            }
+        }
+    }
+
+    if seed_addrs.is_empty() {
+        panic!("Every seed address failed to resolve. Exiting.");
```
Rqnsom, line 114: I guess this should never happen
```suggestion
        panic!("seed address not found");
```

Rqnsom (approved):
I left some ideas, but it looks good - I like it!

src/tools/crawler/main.rs
asmie: Done.

src/tools/crawler/metrics.rs
asmie: Good idea! Changed.

src/tools/crawler/main.rs
asmie: I'm not convinced to killing process if **any ** seed fails. 
I think that there is quite often use case, when eg. in our CI we'll add 10 DNS seeds (every possible seed). Probably, some of them return the same seeds (as each DNS seed can return many addresses) and it would be ok, when 1 out of 10 would be unavailable - I believe we still would like to crawl the network using remaning 9 seeds.
Constantly changing CI configuration because of one or two unavailable seeds seems to generate more workload to maintain the list than just putting all possible seeds and happily ignore if any of them fails.
Besides, user is still notified by `error` message.

src/tools/crawler/README.md
asmie: True, that could be a fine feature to recognize and parse different types of addresses. 
There are, however many use cases to handle:
1. user inputs only IP address (need to add default port);
2. user inputs IP address with port (that's current `seed-addrs` case) but need to parse it (now it's done by clap);
3. user inputs hostname - need to perform DNS lookup and append port to each returned IP;
4. user inputs hostname with port - need to perform DNS lookup and use each returned IP with given port.

Basically, DNS seeder return many addresses during lookup and each address points de facto to different host. Plain host given by name should return only one address but it's possible that it returns more than one address (which could point to the same host) - that wouldn't be desired behavior as we'd like get all A records only for DNS seeder not the plain hosts.

That's something worth considering but I'd do it in separate PR as it may involve a bit more coding.

asmie: Added to improvement bucket 😄

PR #222, opened 2023-04-27
feat: crawler waits for addr message for given timeout

asmie:
This feature changes crawler behavior - till this commit, crawler connected to the peer, grab first addr message and instantly disconnected. Now, crawler waits for the first addr message that contains more than one address OR addr message with one address but different from the peer source address.

Waiting for addr message implies that crawler has to manage connected nodes - it's done using additional field in known_network structure. Connections are managed from the main loop - old connected nodes are disconnected after specified timeout (180 s).

Waiting for addr message makes crawler work slower. There can be observed more "waves" where connections are made and then, some waves where crawler prints it cannot make new connections due to connection limit reached. However, those hosts should be crawler too as they are added to known addresses and not marked as beeing visited.

To avoid error messages about too many connections, connection limit has been slightly increased. Main loop interval has been increased as now crawler works a bit slower, waiting for second addr message for each connection.

### DIFF
--- a/src/tools/crawler/network.rs
+++ b/src/tools/crawler/network.rs
@@ -12,7 +12,7 @@ use ziggurat_core_crawler::connection::KnownConnection;
 pub const LAST_SEEN_CUTOFF: u64 = 10 * 60;
 
 #[derive(Debug, Default, Clone, Copy, PartialEq)]
-pub enum NodeState {
+pub enum ConnectionState {
     /// The node is not connected.
     #[default]
     Disconnected,
@@ -39,7 +39,7 @@ pub struct KnownNode {
     /// The number of subsequent connection errors.
     pub connection_failures: u8,
     /// The node's state.
-    pub state: NodeState,
+    pub state: ConnectionState,
 }
 
 /// The list of nodes and connections the crawler is aware of.
@@ -65,8 +65,8 @@ impl KnownNetwork {
         });
     }
 
-    /// Sets the node's state.
-    pub fn set_node_state(&self, addr: SocketAddr, state: NodeState) {
+    /// Sets the node's connection state.
+    pub fn set_node_state(&self, addr: SocketAddr, state: ConnectionState) {
         if let Some(node) = self.nodes.write().get_mut(&addr) {
             node.state = state;
         }
--- a/src/tools/crawler/network.rs
+++ b/src/tools/crawler/network.rs
@@ -11,6 +11,15 @@ use ziggurat_core_crawler::connection::KnownConnection;
 /// The elapsed time before a connection should be regarded as inactive.
 pub const LAST_SEEN_CUTOFF: u64 = 10 * 60;
 
+#[derive(Debug, Default, Clone, Copy, PartialEq)]
+pub enum NodeState {
+    /// The node is not connected.
+    #[default]
+    Disconnected,
+    /// The node is connected.
+    Connected,
+}
+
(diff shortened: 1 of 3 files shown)


src/tools/crawler/protocol.rs
```diff
@@ -16,11 +16,13 @@ use ziggurat::{
...
 use super::network::KnownNetwork;
+use crate::network::NodeState;
 
 pub const NUM_CONN_ATTEMPTS_PERIODIC: usize = 500;
-pub const MAX_CONCURRENT_CONNECTIONS: u16 = 1000;
-pub const MAIN_LOOP_INTERVAL: u64 = 5;
+pub const MAX_CONCURRENT_CONNECTIONS: u16 = 1200;
+pub const MAIN_LOOP_INTERVAL: u64 = 20;
 pub const RECONNECT_INTERVAL: u64 = 5 * 60;
+pub const MAX_WAIT_FOR_ADDR: u64 = 3 * 60;
```
kylegranger, line 25: For time duration constants, I put the unit in the name of the const.  E.g., `MAIN_LOOP_INTERVAL_SECS`.
Then, I don't have to think about what the scalar applies to.

kylegranger (approved):
👍  very good

src/tools/crawler/network.rs
```diff
@@ -11,6 +11,15 @@ use ziggurat_core_crawler::connection::KnownConnection;
 /// The elapsed time before a connection should be regarded as inactive.
 pub const LAST_SEEN_CUTOFF: u64 = 10 * 60;
 
+#[derive(Debug, Default, Clone, Copy, PartialEq)]
+pub enum NodeState {
```
Rqnsom, line 15: maybe it's a bit more general approach (so it could be reused even, but we won't need it anywhere else here though):
```suggestion
pub enum ConnectionState {
```

src/tools/crawler/network.rs
```diff
@@ -54,6 +65,13 @@ impl KnownNetwork {
         });
     }
 
+    /// Sets the node's state.
```
Rqnsom, line 68: `node state` can mean anything without further context
This could clear things up:
```suggestion
    /// Sets the node's connection state.
```

src/tools/crawler/protocol.rs
```diff
@@ -129,15 +132,26 @@ impl Reading for Crawler {
     async fn process_message(&self, source: SocketAddr, message: Self::Message) -> io::Result<()> {
         match message {
             Message::Addr(addr) => {
-                info!(parent: self.node().span(), "got {} address(es) from {}", addr.addrs.len(), source);
+                let len = addr.addrs.len();
+                info!(parent: self.node().span(), "got {} address(es) from {}", len, source);
```
Rqnsom, line 136: nice!

Rqnsom (approved):
LGTM

src/tools/crawler/protocol.rs
asmie: Done.

src/tools/crawler/network.rs
asmie: You're right - changed.

src/tools/crawler/network.rs
asmie: Changed.

zeapoz (approved):
LGTM :+1:

Rqnsom (approved):
🚀

PR #224, opened 2023-05-02
feat: add synth_node_bin

Rqnsom:
Two things are here:
```
    feat: add synth_node_bin

    Synth node binary can be used to start a synth_node from an external
    binary and connect it to any node via the network protocol.
```
```
    chore: rename package name to a more specific name

    Because the term "ziggurat" is a more general term these days,
    there are multiple ziggurat projects so ZCash implementation of
    the Ziggurat test suite is not unique anymore.
```


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "ziggurat"
+name = "ziggurat-zcash"
```
zeapoz, line 2: Good call changing this!

zeapoz (approved):
LGTM :+1:

synth_node_bin/src/main.rs
```diff
@@ -0,0 +1,125 @@
...
+        fmt()
+            .with_test_writer()
+            .with_env_filter(EnvFilter::from_default_env())
+            .init();
+    }
+
+    loop {
+        println!("Starting a synthetic node.");
+
+        if let Err(e) = start_synth_node(node_addr).await {
```
kylegranger, line 62: I'd call this `run_synth_node` or `do_synth_node` -- you're doing many more things here than starting a synth node.

kylegranger (approved):
🚀  LGTM!

synth_node_bin/src/main.rs
Rqnsom: Good point! Thank you!

Rqnsom: Done

PR #225, opened 2023-05-02
deps: update trivial

zeapoz:
Upgrades trivial dependencies:
```
async-trait 0.1.53 -> 0.1.68
home 0.5.3 -> 0.5.5
metrics 0.20.0 -> 0.21.0
metrics-util 0.15.0
pea2pea 0.45 -> 0.46
toml 0.6.0 -> 0.7.3
clap 4.1.4 -> 4.2.5
```
Also removes the patch part of the SemVer for each package.

*Note:* `tabled` and `jsonrpsee` have new versions available but were not upgraded, since they introduce breaking changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,17 +6,17 @@ edition = "2021"
 
 [dependencies]
 assert_matches = "1.5"
-async-trait = "0.1.53"
+async-trait = "0.1"
 bytes = "1"
 chrono = "0.4"
-dns-lookup = "1.0.8"
-hex = "0.4.3"
-home = "0.5.3"
-lazy_static = "1.4.0"
-metrics = "0.20.0"
-metrics-util = "0.14.0"
+dns-lookup = "2.0"
+hex = "0.4"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -6,30 +6,30 @@ edition = "2021"
 
 [dependencies]
 assert_matches = "1.5"
-async-trait = "0.1.53"
+async-trait = "0.1.68"
```
Rqnsom, line 9: Let's remove the patch part to make it easier from all these deps - the patch version of the SemVer should always have backwards-compatible changes

And it will make crate version maintenance much easier.
```suggestion
async-trait = "0.1"
```

Rqnsom (approved):
LGTM with a comment

Cargo.toml
zeapoz: Good point, done :+1:

Rqnsom (approved):
LGTM 🚀

PR #226, opened 2023-05-02
chore: synth node should use a valid user agent

Rqnsom:
Three small items:
```
    chore: synth node should use a valid user agent

    The synth node should pretend to be a valid node instance,
    so, in this case, let's use MagicBean:5.4.2 and pretend to
    be a zcashd node.
```
```
    chore: change trace for receiving messages

    - trace log level for incoming messages should be `info`
      so it's easier to observe in the logs.
```
```
    chore: add a .gitignore for the synth node bin
```


src/protocol/payload/version.rs
```diff
@@ -51,7 +51,8 @@ impl Version {
                 addr: addr_from,
             },
             nonce: Nonce::default(),
-            user_agent: VarStr(String::from("")),
+            // Let's pretend to be a ZCashd node 5.4.2
+            user_agent: VarStr(String::from("MagicBean:5.4.2")),
```
zeapoz, line 55: Let's make this a constant instead

kylegranger (approved):
LGTM 🚀

zeapoz (approved):
LGTM :+1:

src/protocol/payload/version.rs
Rqnsom: Done

But usually, I follow a rule: if a magic value appears only once, creating a const is not necessary, it's just an overhead

PR #227, opened 2023-05-03
ref: use seed-addrs for both standalone ips and dns

zeapoz:
- IP addresses/DNS servers are now given with the same CLI argument, `--seed-addrs`.
- Crawler will first try to parse an argument as a `SocketAddr`, if that fails it will try a DNS lookup instead.
- Also included are accompanying docs and workflow changes.

### DIFF
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -5,7 +5,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-use clap::{ArgGroup, Parser};
+use clap::Parser;
 use dns_lookup::lookup_host;
 use parking_lot::Mutex;
 use pea2pea::{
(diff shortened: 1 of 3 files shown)


kylegranger (approved):
LGTM

src/tools/crawler/main.rs
```diff
@@ -94,29 +84,25 @@ async fn main() {
...
 
             if let Ok(response) = response {
                 for address in response.iter() {
                     seed_addrs.push(SocketAddr::new(*address, ZCASH_P2P_DEFAULT_PORT)); // DNS addrs use this port
-                    info!(parent: crawler.node().span(), "DNS seed {} address added: {}", seed_str, address);
+                    info!(parent: crawler.node().span(), "DNS seed {} address added: {}", seed_addr, address);
                 }
             } else {
-                error!(parent: crawler.node().span(), "DNS seed {} lookup failed: {}", seed_str, response.err().unwrap());
+                error!(parent: crawler.node().span(), "DNS seed {} lookup failed: {}", seed_addr, response.err().unwrap());
```
Rqnsom, line 101: now well, this could be a general text simply saying "invalid {seed_addr}"

because one might fail to write an address e.g. `1288.100.13.4:8233` and we will print
`DNS seed 1288.100.13.4:8233 lookup failed`
which would be odd, no?

Rqnsom (commented):
LGTM 👍

PR #228, opened 2023-05-05
chore: correctly parse all permutations of ip/dns

zeapoz:
- Crawler can now parse all permutations of inputted `seed-addrs`. These are:
  - IP + port (both IPv4 and IPv6 are valid)
  - IP (Default port will be appended)
  - DNS + Port (Will do a lookup on DNS and remember the port)
  - DNS (Default port will be appended)

### DIFF
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -1,5 +1,5 @@
 use std::{
-    net::SocketAddr,
+    net::{IpAddr, SocketAddr},
     sync::Arc,
     thread,
     time::{Duration, Instant},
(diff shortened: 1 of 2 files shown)


src/tools/crawler/main.rs
```diff
@@ -75,33 +75,68 @@ fn start_logger(default_level: LevelFilter) {
...
+/// Parses and converts `String` values found in a `Vec` to valid `SocketAddr`.
+///
+/// # Input
+///
+/// Valid inputs can be in the following forms:
+/// - IP + port (both IPv4 and IPv6 are valid)
+/// - IP (Default `ZCASH_P2P_DEFAULT_PORT` will be appended)
+/// - DNS + port
+/// - DNS
+fn parse_addrs(seed_addrs: Vec<String>, crawler: &Crawler) -> Vec<SocketAddr> {
```
Rqnsom, line 87: a unit test could be written for something like this - we should not forget to do that sometimes, from time to time 😃

src/tools/crawler/main.rs
```diff
@@ -75,33 +75,68 @@ fn start_logger(default_level: LevelFilter) {
...
 }
 
+/// Parses and converts `String` values found in a `Vec` to valid `SocketAddr`.
+///
+/// # Input
+///
+/// Valid inputs can be in the following forms:
+/// - IP + port (both IPv4 and IPv6 are valid)
+/// - IP (Default `ZCASH_P2P_DEFAULT_PORT` will be appended)
+/// - DNS + port
```
asmie, line 85:
```suggestion
/// - hostname (can be DNS seeder) + port
```

src/tools/crawler/main.rs
```diff
@@ -75,33 +75,68 @@ fn start_logger(default_level: LevelFilter) {
...
+/// - DNS
```
asmie, line 86:
```suggestion
/// - hostname (can be DNS seeder, default `ZCASH_P2P_DEFAULT_PORT` will be appended)
```

src/tools/crawler/main.rs
```diff
@@ -75,33 +75,68 @@ fn start_logger(default_level: LevelFilter) {
...
+        // First, try parsing as a `SocketAddr`.
+        if let Ok(addr) = seed_addr.parse::<SocketAddr>() {
+            parsed_addrs.push(addr);
+            continue;
+        }
+        // User may supply an IP address without a port,
+        // append `ZCASH_P2P_DEFAULT_PORT` in that case.
+        if let Ok(addr) = seed_addr.parse::<IpAddr>() {
+            parsed_addrs.push(SocketAddr::new(addr, ZCASH_P2P_DEFAULT_PORT));
+            info!(parent: crawler.node().span(), "no port specified for address: {}, using default: {}", seed_addr, ZCASH_P2P_DEFAULT_PORT);
```
Rqnsom, line 100: would this be more applicible approach for an improved command output?
```suggestion
            println!(parent: crawler.node().span(), "no port specified for address: {}, using default: {}", seed_addr, ZCASH_P2P_DEFAULT_PORT);
```

src/tools/crawler/main.rs
```diff
@@ -75,33 +75,68 @@ fn start_logger(default_level: LevelFilter) {
...
+            continue;
+        }
+        // If above failed, try to do a DNS lookup instead.
+        //
+        // We make sure to remove remove the port if it exists.
```
Rqnsom, line 105: 🤔
maybe there will be out there some DNS addresses with a non-default port and we won't allow the user to crawl those addresses

we can keep it simple - append a port if it is not provided

asmie (approved):
Very nice! 
It should cover now all the cases :)

Rqnsom (approved):
> fix: correctly parse all permutations of ip/dns

It's not a `fix` unless we have a bug. this feels more like a `chore` since it's an improvement

LGTM 🚀

src/tools/crawler/main.rs
asmie: Port is extracted and added later, when hostname is resolved to IPaddr - so the port is preserved as the user specified.

Rqnsom: aaa, I see! That's why some unit test input examples would be nice to have here - it would quickly clarify something like this

src/tools/crawler/main.rs
zeapoz: I'll add some tests for the parsing of standard IP addresses for now. Not sure if we could add tests for hostname resolution since that depends on external factors :thinking:

src/tools/crawler/main.rs
zeapoz: Done

src/tools/crawler/main.rs
zeapoz: Done

src/tools/crawler/main.rs
zeapoz: Done

src/tools/crawler/main.rs
zeapoz: Done

Rqnsom (approved):

PR #229, opened 2023-05-12
feat: alert message type

zeapoz:
Synth node may sometimes fail because of alert messages sent by ancient real nodes. Since alert messages have been deprecated since long ago, we ignore them.


Rqnsom (approved):
LGTM

PR #230, opened 2023-05-15
ci: remove obsolete line

Rqnsom:
We don't run zcashd in the background so we don't need to kill it after the crawling is over.


zeapoz (approved):
Good catch! LGTM 👌

PR #231, opened 2023-05-15
 feat: support for WtxMxg type in InvHash 

Rqnsom:
Ensures that our synth node binary (or a crawler) won't die within a few seconds after performing a handshake with zebrad


asmie (approved):
LGTM

PR #232, opened 2023-05-17
chore: update synth_node_bin's default action 

Rqnsom:
Updated synth_node_bin default action to run the synth node indefinitely until the connection is dropped.
The new version of the synth node is resistant to panicking, so when the `stubborn` argument is used, it will run forever.

For this purpose, in the synth_node.rs a new API has been introduced which won't panic in case of failure but will return an `Err` instead:
```rust
    /// Reads a message from the inbound (internal) queue of the node.
    /// In case of channel failure, it returns an [`ConnectionError`].
    ///
    /// Messages are sent to the queue when unfiltered by the message filter.
    pub async fn try_recv_message(&mut self) -> Result<(SocketAddr, Message), ConnectionError>
```

### DIFF
--- a/synth_node_bin/src/main.rs
+++ b/synth_node_bin/src/main.rs
@@ -85,7 +85,7 @@ async fn run_synth_node(node_addr: SocketAddr) -> Result<()> {
     synth_node.connect(node_addr).await?;
 
     // Run the wanted action with the node.
-    perform_action(&mut synth_node).await;
+    perform_action(&mut synth_node, node_addr).await?;
 
     // Optional.
     sleep(Duration::from_millis(100)).await;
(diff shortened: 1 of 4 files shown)


synth_node_bin/src/main.rs
```diff
@@ -102,24 +102,20 @@ async fn run_synth_node(node_addr: SocketAddr) -> Result<()> {
 //
 // All the program logic happens here.
 #[allow(unused_variables)]
-async fn perform_action(synth_node: &mut SyntheticNode) {
+async fn perform_action(synth_node: &mut SyntheticNode, addr: SocketAddr) {
```
asmie, line 105: Just thinking if `perform_action` should return `Result<()>` as `run_synth_node` returns that. 
Now we've a path, where if `unicast` fails, there will be warning message displayed, `run_synth_node` will go further and return `Ok(())` to the `main` instead of `Err`. So we'll be returning `Ok(())` even if there was something wrong.

For now it doesn't matter but maybe at some point in the future we'll use it more extensively.

asmie (approved):
Nice 👍 
I always like when application tries to protect and recover itself and not gonna die quickly 😄

synth_node_bin/src/main.rs
Rqnsom: Done!
For now, our binary shouldn't panic easily, let's monitor it!

asmie (approved):
👍

PR #233, opened 2023-05-19
ci: run crawler on self-hosted machine

zeapoz:
Also adds `workflow_dipsatch` so we can manually trigger the workflow.


Rqnsom (approved):
Do you need approval?

PR #234, opened 2023-05-19
Call ziggurat-core:build-viz-state workflow after crawl.

muzam1l:
Related to https://github.com/runziggurat/ziggurat-core/pull/41

### DIFF
--- a/.github/workflows/crawler.yml
+++ b/.github/workflows/crawler.yml
@@ -28,9 +28,13 @@ jobs:
         with:
           name: latest-result
           path: latest.json
+          
+  call-build-viz-state-workflow:
+    needs: [ crawl-network ]
+    uses: runziggurat/ziggurat-core/.github/workflows/build-viz-state.yml@main
 
   call-process-results-workflow:
-    needs: [ crawl-network ]
+    needs: [ call-build-viz-state-workflow ]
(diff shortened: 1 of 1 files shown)


.github/workflows/crawler.yml
zeapoz, line 1: We're uploading two files as `latest-result` (`crawler` and the `viz-state`). Will the second one overwrite the first one, which would then get uploaded under `crawler/`, or does it include both?

muzam1l: `latest-result` is an artifact name, more like a bucket/top-level-folder in which the files go. We upload `latest.json` here and then [build-viz-state.yml](https://github.com/runziggurat/ziggurat-core/pull/41/files/590f949f988e612cf1e55b3353798f62f2b06eaf#diff-231e35c35d4b8139338623a0eca254593697ce3c66dfc46aac2b621a4e0f9967) uploads `latest.viz.json` to it. Then they are both uploaded by `process-results.yml` to gcs, along with gzipped copy of `latest.json` (`latest.viz.json` is not gzip-copied).

muzam1l: Had some confusions about it too, so created a [demo](https://github.com/muzam1l/test-artifact-actions/actions/runs/5006490579)!

zeapoz (approved):

PR #239, opened 2023-05-23
chore: wait for Version message before sending GetAddr

asmie:
During tests some of the peers have been observed to have delays (massive!) in sending first Version message.
Till that commit crawler:
1) had degenerated handshake procedure - it was waiting only for crawler to send version and returned success;
2) had hardcoded 1 sec sleep after successfull connect() and after that it just sent GetAddr message.

The above resulted in situations, where GetAddr was sent before the remote party sent crawler version message. It leads to dropping GetAddr message as the protocol standard says that there can be no other communication before exchanging Version messages.

This commit removes hardcodes and make crawler fit into protocol standard. Now GetAddr messages are sent only after crawler receives Version message from the remote party. That allows to wait even up to crawler connection limit (120 secs for single connection).

### DIFF
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -168,9 +168,8 @@ async fn main() {
                 .write()
                 .insert(addr, KnownNode::default());
 
-            // Don't care about the result - errors are logged inside connect() and
-            // successful connections do not involve any further actions as sending
-            // GetAddr message was moved to process_message in protocol.rs.
+            // Once the Version message is received in the process_message function,
+            // GetAddr will be requested from the peer
             let _ = crawler_clone.connect(addr).await;
         });
     }
@@ -232,9 +231,8 @@ async fn main() {
                 if crawler.should_connect(addr) {
                     let crawler_clone = crawler.clone();
                     tokio::spawn(async move {
-                        // Don't care about the result - errors are logged inside connect() and
-                        // successful connections do not involve any further actions as sending
-                        // GetAddr message was moved to process_message in protocol.rs.
+                        // Once the Version message is received in the process_message function,
+                        // GetAddr will be requested from the peer
                         let _ = crawler_clone.connect(addr).await;
                     });
                 }
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -17,7 +17,7 @@ use tokio::{signal, time::sleep};
 use tracing::{debug, error, info, warn};
 use tracing_subscriber::filter::{EnvFilter, LevelFilter};
 use ziggurat_core_crawler::summary::NetworkSummary;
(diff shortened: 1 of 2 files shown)


zeapoz (approved):
LGTM :+1:

src/tools/crawler/protocol.rs
```diff
@@ -177,6 +181,17 @@ impl Reading for Crawler {
...
+
+                // Send GetAddr as soon as we get version message from the peer.
+                // In fact, this part should be done during the handshake but it would increase
+                // handshake time and there are some nodes that do not send version message
+                // quickly (we know that zebra can delay sending version message for over 30 seconds).
+                // Sending GetAddr before receiving version results in dropping this message by
+                // the remote peer so we're stuck waiting for reply that will never come that's why we
+                // need to wait for remote version message.
+                // Sending GetAddr message was moved to this place to and it's not sent anymore
+                // directly from main module.
```
Rqnsom, line 193:
```suggestion
                // Send GetAddr as soon as we get version message from the peer.
                // In fact, this part should be done during the handshake but it would increase
                // handshake time and there are some nodes that do not send version message
                // quickly (we know that zebra can delay sending version message for over 30 seconds).
                // Sending GetAddr before receiving the version results in dropping this message by
                // the remote peer, so we're stuck waiting for a reply that will never come that's why we
                // need to wait for the remote version message response.
                // Extra background: Sending GetAddr message was moved to this place,
                // and it's not sent anymore directly from the main module.
```

src/tools/crawler/main.rs
```diff
@@ -232,16 +232,14 @@ async fn main() {
...
                     let crawler_clone = crawler.clone();
                     tokio::spawn(async move {
-                        if crawler_clone.connect(addr).await.is_ok() {
-                            sleep(Duration::from_secs(1)).await;
-                            let _ = crawler_clone.unicast(addr, Message::GetAddr);
-                        }
+                        // Don't care about the result - errors are logged inside connect() and
+                        // successful connections do not involve any further actions as sending
+                        // GetAddr message was moved to process_message in protocol.rs.
+                        let _ = crawler_clone.connect(addr).await;
```
Rqnsom, line 238: We don't need to document historical reasons, we can just point out where the new solution is and how it works shortly
```suggestion
                        // Once the Version message is received in the process_message function,
                        // GetAddr will be requested from the peer
                        let _ = crawler_clone.connect(addr).await;
```

Rqnsom (approved):
LGTM

src/tools/crawler/main.rs
asmie: Changed.

src/tools/crawler/protocol.rs
asmie: Changed.

PR #242, opened 2023-05-29
feat: add default_port argument to crawler command line

asmie:
Add new command line argument which allows user to specify default port to be used with seeds with no port specified. It allows running crawler both with mainnet and testnet as well as nodes running on non-standard ports.

Note: there is still a need to change network MAGIC value to switch between mainnet and testnet.

### DIFF
--- a/src/tools/crawler/README.md
+++ b/src/tools/crawler/README.md
@@ -22,6 +22,9 @@ OPTIONS:
     -s, --seed-addrs <SEED_ADDRS>...
             A list of initial standalone IP addresses and/or DNS servers to connect to
 
+    -n, --node-listening-port <NODE_LISTENING_PORT>
+            Default port used for connecting to the nodes [default: 8233]
+
     -V, --version
             Print version information
 ```
--- a/src/tools/crawler/main.rs
+++ b/src/tools/crawler/main.rs
@@ -56,7 +56,7 @@ struct Args {
 
     /// Default port used for connecting to the nodes
     #[clap(short, long, value_parser, default_value_t = ZCASH_P2P_DEFAULT_MAINNET_PORT)]
-    default_port: u16,
+    node_listening_port: u16,
     // TODO
     // #[clap(short, long, value_parser, default_value = "testnet")]
     // network: String,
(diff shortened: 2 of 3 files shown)


src/tools/crawler/main.rs
```diff
@@ -53,6 +53,10 @@ struct Args {
     /// If present, start an RPC server at the specified address
     #[clap(short, long, value_parser)]
     rpc_addr: Option<SocketAddr>,
+
+    /// Default port used for connecting to the nodes
+    #[clap(short, long, value_parser, default_value_t = ZCASH_P2P_DEFAULT_MAINNET_PORT)]
+    default_port: u16,
```
Rqnsom, line 59: maybe the logic can be: 

default ports are 8233 and 18233 by definition, which we assign here to a listening port
```suggestion
    /// Listening port used for connecting to the nodes for which the port is unknown.
    /// If not provided, a default port for the mainnet/testnet network will be used.
    #[clap(short, long, value_parser, default_value_t = ZCASH_P2P_DEFAULT_MAINNET_PORT)]
    listening_port: u16,
```

Rqnsom (approved):
> Note: there is still a need to change network MAGIC value to switch between mainnet and testnet.

Maybe we can introduce in some future PR a cargo build flag which can be used to pick a wanted MAGIC value.

LGTM with a small suggestion! 🚀

src/tools/crawler/main.rs
asmie: Maybe `node_listening_port` to not confuse anyone that it's our listening port?

asmie:
It can be configured right now using `test` and `crawler`  cargo flags.

```
#[cfg(test)]
pub const MAGIC: [u8; MAGIC_LEN] = MAGIC_TESTNET;
#[cfg(all(not(test), not(feature = "crawler")))]
pub const MAGIC: [u8; MAGIC_LEN] = MAGIC_MAINNET;
#[cfg(all(not(test), feature = "crawler"))]
pub const MAGIC: [u8; MAGIC_LEN] = MAGIC_MAINNET;
```

I've got another idea - ideally if it would be configurable during runtime instead of compile-time. Probably, it could be solved by global singleton object returning proper magic value for all messages. It could even allow to start sending messages with one magic and then switch to another (in the middle of communication - just curious how zcash daemons would react 😄 ). But that should be separate PR as it will be change in the core files, may affect also ziggurat test suite.

src/tools/crawler/main.rs
Rqnsom: sounds even better!

Rqnsom:
> It could even allow to start sending messages with one magic and then switch to another (in the middle of communication - just curious how zcash daemons would react 😄 ).

Uuu yes! I would love to try that

src/tools/crawler/main.rs
asmie: Done

src/tools/crawler/README.md
```diff
@@ -22,6 +22,9 @@ OPTIONS:
     -s, --seed-addrs <SEED_ADDRS>...
             A list of initial standalone IP addresses and/or DNS servers to connect to
 
+    -n, --node-listening-port <NODE_LISTENING_PORT>
+            Default port used for connecting to the nodes [default: 8233]
```
Rqnsom, line 26: yes!

Rqnsom (approved):

PR #243, opened 2023-06-01
refactor(synth_node_bin): support multiple actions

Rqnsom:
A synth node binary should have an easy way to add new actions. This implementation provides action plugin support so that all different actions can be added under different modules under the `src/action/` directory.

### DIFF
--- a/synth_node_bin/src/action/mod.rs
+++ b/synth_node_bin/src/action/mod.rs
@@ -1,0 +1,56 @@
+use std::net::SocketAddr;
+
+use anyhow::Result;
+use ziggurat_zcash::tools::synthetic_node::SyntheticNode;
+
+mod send_get_addr_and_forever_sleep;
+
+/// Defines properties of any action for a synth node binary.
+///
+/// It simplifies adding new actions and allows to separate different actions with modules.
+#[async_trait::async_trait]
+trait SynthNodeAction {
+    /// Action description.
+    ///
+    /// It can be displayed during the runtime.
+    fn info(&self) -> &str;
+
+    /// Defines the core action functionality.
+    ///
+    /// All the program logic happens here.
+    async fn run(&self, synth_node: &mut SyntheticNode, addr: SocketAddr) -> Result<()>;
+}
+
+/// List of available actions.
+pub enum ActionType {
+    SendGetAddrAndForeverSleep,
+}
+
+/// Action handler.
+pub struct ActionHandler {
+    /// Internal action.
+    action: Box<dyn SynthNodeAction>,
+}
+
+impl ActionHandler {
+    /// Creates a new [`ActionHandler`] for a given [`ActionType`].
+    pub fn new(action_type: ActionType) -> Self {
+        Self {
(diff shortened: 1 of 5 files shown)


synth_node_bin/src/action/send_get_addr_and_forever_sleep.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[async_trait::async_trait]
+impl SynthNodeAction for Action {
+    fn info(&self) -> &str {
+        "request an GetAddr and then sleeps forever"
+    }
+
+    #[allow(unused_variables)]
+    async fn run(&self, synth_node: &mut SyntheticNode, addr: SocketAddr) -> Result<()> {
+        println!("Synthetic node performs an action.");
```
asmie, line 19: Maybe we should focus on one type of logging/printing. If we're using `tracing::` so maybe we'd convert that to `tracing::info!`

synth_node_bin/src/action/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+                ActionType::SendGetAddrAndForeverSleep => {
+                    send_get_addr_and_forever_sleep::Action {}
+                }
+            }),
+        }
+    }
+
+    /// Runs the underlying action.
+    pub async fn execute(&self, synth_node: &mut SyntheticNode, addr: SocketAddr) -> Result<()> {
+        println!(
```
asmie, line 49: Maybe `tracing::info`?

synth_node_bin/src/action/send_get_addr_and_forever_sleep.rs
Rqnsom: I think the binary should have a stdout printout as well in case the RUST_LOG is not used.

So the most important info - like action info, was the synt node started/restarted could be printed with `println`.
Because I don't always start it with `RUST_LOG`

asmie (commented):
Very nice change! 👍 
I like the idea of creating catalogue of actions - over some time we can make a nice collection!

Wondering if it would be possible to switch between those actions in runtime?
Shouldn't be there introduced some generic mechanism to choose actions using command line? Did you think about it?

synth_node_bin/src/action/send_get_addr_and_forever_sleep.rs
asmie: Convinced me 😸

Rqnsom (commented):
> Wondering if it would be possible to switch between those actions in runtime?
Shouldn't be there introduced some generic mechanism to choose actions using command line? Did you think about it?

Yeah, that was the initial plan! Not during the runtime, but before starting the binary - we should be able to choose which action to use, check this old TODO comment copied from xrpl:

https://github.com/runziggurat/zcash/blob/6afeacd0b92e6e2922bf890616f1aac46c6af87b/synth_node_bin/src/main.rs#L10

Could you approve? 😃

asmie (approved):
Good, so there should be a task in backlog to continue working over adding commands for steering actions from CLI.

PR #244, opened 2023-06-02
feat(synth_node_bin): add advanced SN for s001

Rqnsom:
Commits with info about the change:
```
    feat(synth_node_bin): add advanced SN for s001

    A synth_node_bin with this action can:
     - listen for inbound connections and print connection status periodically
     - periodically send GetAddr/Ping/Addr messages to all connected peers
```
```
    feat(synthetic_node): add connected_peer_infos API
```
```
    feat(synth_node_bin): support for action preconfig

    - It should be possible to configure the action settings before the
      action is even started in some cases.
```

There's one more TODO in the code regarding the listening address, which needs to be fed to the synth_node builder - which will be done in a separate PR.

### DIFF
--- a/synth_node_bin/src/action/advanced_sn_for_s001.rs
+++ b/synth_node_bin/src/action/advanced_sn_for_s001.rs
@@ -1,0 +1,183 @@
+use std::{net::SocketAddr, str::FromStr};
+
+use anyhow::Result;
+use tokio::time::{interval, sleep, Duration};
+use ziggurat_zcash::{
+    protocol::{
+        message::Message,
+        payload::{addr::NetworkAddr, Addr, Nonce},
+    },
+    tools::{
+        message_filter::{Filter, MessageFilter},
+        synthetic_node::SyntheticNode,
+    },
+};
+
+use super::{ActionCfg, SynthNodeAction};
+
+// Configurable status printout interval.
+const DBG_INFO_LOG_INTERVAL_SEC: Duration = Duration::from_secs(10);
+const BROADCAST_INTERVAL_SEC: Duration = Duration::from_secs(120);
+
+pub(super) struct Action;
+
+pub(super) fn action() -> Box<dyn SynthNodeAction> {
+    Box::new(Action {})
+}
+
+#[async_trait::async_trait]
+impl SynthNodeAction for Action {
+    fn info(&self) -> &str {
+        "an advanced SN node created for RT-S001 which can do the following
+           - listens for inbound connections and prints the connection status periodically
+           - periodically requests and sends Ping/GetAddr/Addr messages to all peers"
+    }
+
+    fn config(&self) -> ActionCfg {
+        ActionCfg {
+            msg_filter: MessageFilter::with_all_auto_reply().with_getaddr_filter(Filter::Disabled),
+        }
+    }
+
+    #[allow(unused_variables)]
+    async fn run(&self, synth_node: &mut SyntheticNode, addr: SocketAddr) -> Result<()> {
+        // Sleep for three seconds before taking any actions - so the GetHeaders is handled before we
+        // send GetAddr to the zcashd node for our only outbound connection.
+        sleep(Duration::from_secs(3)).await;
+
+        let msg = Message::GetAddr;
+        tracing::info!("unicast {msg:?}\n");
+        if synth_node.unicast(addr, msg.clone()).is_err() {
+            tracing::warn!("failed to send {msg:?}\n");
+            anyhow::bail!("connection closed");
+        }
+
+        let mut dbg_info_interval = interval(DBG_INFO_LOG_INTERVAL_SEC);
+        let mut broadcast_msgs_interval = interval(BROADCAST_INTERVAL_SEC);
+
+        loop {
+            tokio::select! {
+                _ = dbg_info_interval.tick() => {
+                    trace_debug_info(synth_node).await;
+                },
+                _ = broadcast_msgs_interval.tick() => {
+                    let _ = broadcast_periodic_msgs(synth_node);
+                },
+                Ok((src, msg)) = synth_node.try_recv_message() => {
+                    handle_rx_msg(synth_node, src, msg).await?;
+                },
+            }
+        }
+    }
+}
+
+async fn trace_debug_info(synth_node: &SyntheticNode) {
+    let peer_infos = synth_node.connected_peer_infos();
+    let peer_cnt = peer_infos.len();
+
+    let mut log = format!("\nNumber of peers: {peer_cnt}\n");
+
+    // Let's sort by the connection's time value.
+    let mut peer_infos: Vec<_> = peer_infos.iter().collect();
+    peer_infos.sort_by(|a, b| a.1.stats().created().cmp(&b.1.stats().created()));
+
+    for (addr, info) in peer_infos.iter() {
+        let stats = info.stats();
+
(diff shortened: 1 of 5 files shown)


synth_node_bin/src/action/mod.rs
```diff
@@ -15,42 +16,72 @@ trait SynthNodeAction {
...
     /// All the program logic happens here.
     async fn run(&self, synth_node: &mut SyntheticNode, addr: SocketAddr) -> Result<()>;
 }
 
 /// List of available actions.
+// TODO: Make a command argument to choose action type.
+#[allow(dead_code)]
 pub enum ActionType {
     SendGetAddrAndForeverSleep,
+    // TODO(Rqnsom): Add support for choosing listening address in config and apply it in the main.rs here. Details:
```
asmie, line 35: What are the real-life scenarios where we'd like to narrow listening to some specified IP instead of 0.0.0.0 (all) for synth node binary?
Is it worth time spent?

synth_node_bin/src/action/advanced_sn_for_s001.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    fn config(&self) -> ActionCfg {
+        ActionCfg {
+            msg_filter: MessageFilter::with_all_auto_reply().with_getaddr_filter(Filter::Disabled),
+        }
+    }
+
+    #[allow(unused_variables)]
+    async fn run(&self, synth_node: &mut SyntheticNode, addr: SocketAddr) -> Result<()> {
+        // Sleep for 3 seconds before taking any actions.
```
asmie, line 44: What's the reason for slipping 3 secs? 
Now, this comment just duplicate the code and state the obvious, telling just what the next line is doing.
Maybe it would make the comment meaningful when it could include some justification about the chosen delay here.

synth_node_bin/src/action/advanced_sn_for_s001.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    // Let's sort by the connection's time value.
+    let mut peer_infos: Vec<_> = peer_infos.iter().collect();
+    peer_infos.sort_by(|a, b| a.1.stats().created().cmp(&b.1.stats().created()));
+
+    for (addr, info) in peer_infos.iter() {
+        let stats = info.stats();
+
+        // Align all possible IP addresses (both v4 and v6) vertically
+        const MAX_IPV6_ADDR_LEN: usize = 37;
```
asmie, line 87: Why 37? 
8 groups of 4 digits with 7 `:` between them can give 39.

There are possible some other cases like [IPv4-mapped IPv6 address](https://www.rfc-editor.org/rfc/rfc4291#section-2.5.5.2) where max length would be 45. That's why INET6_ADDRSTRLEN constant in Linux has 46 bytes (45 + 1 null terminating char).

synth_node_bin/src/action/advanced_sn_for_s001.rs
```diff
@@ -0,0 +1,181 @@
...
+        if synth_node.unicast(addr, msg.clone()).is_err() {
+            tracing::warn!("failed to send {msg:?}\n");
+            anyhow::bail!("connection closed");
+        }
+
+        let mut dbg_info_interval = interval(DBG_INFO_LOG_INTERVAL_SEC);
+        let mut broadcast_msgs_interval = interval(BROADCAST_INTERVAL_SEC);
+
+        loop {
+            tokio::select! {
```
asmie, line 58: Like it 👍

synth_node_bin/src/action/advanced_sn_for_s001.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+async fn handle_rx_msg(
+    synth_node: &mut SyntheticNode,
+    src: SocketAddr,
+    msg: Message,
+) -> Result<()> {
+    tracing::info!("message received from {src}:\n{msg:?}");
+
+    // We are only handling a GetAddr for now.
```
asmie, line 146: Shouldn't we also handle `ping` messages to ensure other hosts that connection is valid?

synth_node_bin/src/action/advanced_sn_for_s001.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    Ok(())
+}
+
+fn unicast_addr_msg(synth_node: &mut SyntheticNode, dst: SocketAddr) -> Result<()> {
+    // These are all our IPs
+    let addrs = vec![
+        // Our zebrad
+        NetworkAddr::new(SocketAddr::from_str("35.210.208.185:8233").unwrap()),
```
asmie, line 158: It could be configurable in the future to easily change advertised addresses. It could be useful to switch between testnet and mainnet addresses.
TBH I find it could be more useful to change those list from configuration file than setting a listening IP address (but it's only my feeling) 😃

synth_node_bin/src/action/mod.rs
Rqnsom: Yes. But because the default choice in our code is LOCALHOST for the test suite, and the synth node binary mostly uses 0.0.0.0 ----> so we should be able to change it automatically, not doing manual code edits

synth_node_bin/src/action/advanced_sn_for_s001.rs
Rqnsom: There are automatic auto-reply filters applied before the messages reach the inbound msg queue in the `try_recv_message` API, the only autoreply filter which we don't use here is for `GetAddr` messages so we can handle those manually here

Check line 38:
```rust
msg_filter: MessageFilter::with_all_auto_reply().with_getaddr_filter(Filter::Disabled),
```

synth_node_bin/src/action/advanced_sn_for_s001.rs
Rqnsom: > 8 groups of 4 digits with 7 : between them can give 39.

ops, my miscalculation here haha

> There are possible some other cases like [IPv4-mapped IPv6 address](https://www.rfc-editor.org/rfc/rfc4291#section-2.5.5.2) where max length would be 45. That's why INET6_ADDRSTRLEN constant in Linux has 46 bytes (45 + 1 null terminating char).

TIL about this - okay, then let's use 46 here

synth_node_bin/src/action/advanced_sn_for_s001.rs
Rqnsom: for now it's hardcoded, I didn't wanna fit too much stuff in this PR - will be more flexible in the future

synth_node_bin/src/action/advanced_sn_for_s001.rs
Rqnsom: Done

synth_node_bin/src/action/advanced_sn_for_s001.rs
Rqnsom: Done

synth_node_bin/src/action/advanced_sn_for_s001.rs
asmie: Got it  :)

synth_node_bin/src/action/advanced_sn_for_s001.rs
asmie: Great to hear! 👍

asmie (approved):
LGTM 👍

PR #246, opened 2023-06-05
feat(synth_node_bin): add network cfg to ActionCfg 

Rqnsom:
Introducing an option to configure NodeConfig while setting up the synth node with a SynthNodeBuilder.
So that a user could choose a listening address for a synth node implementation.

Now the AdvancedSnForS001 is fully ready to be used.
Two commits:
```
    feat(synth_node_bin): add network cfg to ActionCfg
```
```
    feat(synth_node): support for flexible network config
```


asmie (approved):
LGTM 👍

PR #249, opened 2023-06-06
feat(synth_node_bin): new action for quick connect/disconnect

Rqnsom:
The action quickly connects and then cleanly disconnects.
The word `clean` is used here because one more similar action will be added where we will skip the disconnect API on the synth node side.


asmie (approved):
LGTM 👍

PR #251, opened 2023-06-07
feat(synth_node_bin): add arg for action selection

Rqnsom:
With this change the action can be chosen during the startup of the process.

New help text output:
```
A synthetic node which can connect to the node and preform some actions independently

Usage: synth_node_bin [OPTIONS]

Options:
  -n, --node-addr <NODE_ADDR>      An address of the node in the <ip>:<port> format
  -s, --stubborn                   Always reconnect in the case the connection fails - synthetic node never dies
  -t, --tracing                    Enable tracing
  -a, --action-type <ACTION_TYPE>  Possible actions: SendGetAddrAndForeverSleep / AdvancedSnForS001 / QuickConnectAndThenCleanDisconnect [default: SendGetAddrAndForeverSleep]
  -h, --help                       Print help
  -V, --version                    Print version
```

### DIFF
--- a/synth_node_bin/src/main.rs
+++ b/synth_node_bin/src/main.rs
@@ -1,15 +1,5 @@
 //! A synthetic node binary can be used to interact with the node in the
 //! background from a different runtime environment.
-//!
-//! This is a quick solution - in future Ziggurat projects it will be done differently.
-//!
-//! Remaining worklist (probably not going to be implemented in this repo):
-//!   - Add an argument option to choose specific action in order to support
-//!     different synthetic node binary implementations at once. A few examples:
-//!     ```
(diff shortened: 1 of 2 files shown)


synth_node_bin/src/main.rs
```diff
@@ -1,31 +1,22 @@
...
 use action::{ActionHandler, ActionType};
 use anyhow::Result;
 use clap::Parser;
 use ziggurat_zcash::tools::synthetic_node::SyntheticNode;
 
+use crate::ActionType::SendGetAddrAndForeverSleep;
+
 mod action;
 
-/// A synthetic node which can connect to the XRPL node and preform some
```
Rqnsom, line 22: "XRPL" 😃

PR #252, opened 2023-06-07
feat: add new action which constantly ask for random blocks using getdata command

asmie:
Intended to be used with file with block hashes (JSON file) grabbed either from mainnet or testnet.


Rqnsom (approved):
LGTM :+1:

PR #253, opened 2023-06-08
feat(synth_node_bin): new variant for the connect/disconnect action

Rqnsom:
This new action is pretty similar to the existing one, but in this case, the socket is not closed properly.

commits:
```
    feat(synth_node_bin): new variant of connect/disconnect action

    The action quickly connects and then disconnects without closing the
    socket properly.
```
```
    feat(synth_node_bin): action cfg allow_shutdown option
```


asmie (approved):
LGTM 👍

PR #255, opened 2023-06-12
ci: `chmod +x zebrad`

zeapoz:


muzam1l (approved):

zeapoz:
> Maybe the actions should just fail on these errors instead of producing empty results file?

Ideally yes, I'll look into this, but I think this might be hard to implement. The main issue is that we use `continue-on-error` on the suite step (since `cargo test` may exit with a non-zero exit code). I'm not sure if we can somehow detect if it's due to an external issue or something else, at least I think this might be hard to do in the CI, but I'll investigate a little.

PR #256, opened 2023-06-12
feat(synth_node_bin): Add collector action 

Rqnsom:
```
    chore: make synthetic node more robust
```
```
    refactor(synth_node_bin): make address arg optional
```
```
    feat(synth_node_bin): Add collector action

    A node which sends an Addr message every three seconds containing all
    connected peers and their listening ports.
```


asmie (approved):
LGTM 👍

PR #258, opened 2023-06-13
 feat(synth_node_bin): add rt S1 tainter action 

Rqnsom:
```
    feat(synth_node_bin): add desired_listening_port arg
```
```
    feat(synth_node_bin): add rt S1 tainter action

    - a listener node which sends a tainted Addr message to all its peers
```


asmie (approved):
👍

PR #266, opened 2026-07-23
Update protocol version from 170150 to 170160

autotunafish:
NU6.3
Activation Height | 3,428,143
Expected Date | ~2026-07-28
Consensus Branch ID | 0x37a5165b
Testnet Activation | Block 4,134,000 (active since v6.0.0-rc.0)

And the protocol version 170160. I double-checked with an engineer, I'll try to find the spec that declares it.


aphelionz:
Thanks!

